PR #9, opened 2025-05-06
fix: simplify distribution validation rule

leseb:
The previous validation rule for DistributionType was overly complex and required explicit empty values for unused fields. This change:

- Replaces the complex validation rule with a simpler `has()` check
- Ensures mutual exclusivity between name and image fields

The new validation rule `has(self.name) != has(self.image)` more clearly expresses the intent that only one of the fields should be present.


leseb:
Missed https://github.com/llamastack/llama-stack-k8s-operator/pull/7

PR #16, opened 2025-05-06
ci: introduce a pre-commit check and contributing doc

leseb:
This CI check replaces the linter action, it has been included into the pre-commit hook.
The hook has 2 specialized ones:

* the go linter
* the generate manifest artifacts

You can run pre-commit locally:

```
pre-commit run --all-files
```

You can also install it so it will run before committing code.

```
pre-commit install
```


leseb:
I think there is a back and forth between our local golang versions. @VaishnaviHire

config/rbac/role.yaml
```diff
@@ -2,67 +2,25 @@
...
-  - ""
-  resources:
   - pods/log
   verbs:
   - '*'
 - apiGroups:
   - ""
   resources:
   - serviceaccounts
+  - services/finalizers
```
rhuss, line 23: What is generating this file ? It looks quite random what resources are handled together and when `*` is used or an explicit list of verbs. E.g services/finalizers and services have the same list of verbs, but defined in different sections, and one has a conflated list of verbs, the other uses a wildcard.

Should we manually curate this role definition file maybe ?

VaishnaviHire: Its generated from these annotations [here](https://github.com/llamastack/llama-stack-k8s-operator/blob/main/controllers/kubebuilder_rbac.go)

leseb:
> Looks good to me. Could you briefly elaborate the benefit of pre-commit in addition to `make lint` as this is an additional development dependency that needs to be installed ?

As explained in the CONTRIBUTING guide, installing pre-commit locally is optional, it's to the developers.

> 
> Another idea, would to describe in CONTRIBUTING.md how to install it as a local pre-commit hook for git, so that it is automatically called on a `git commit` (or is this too much and we want to keep more control when to run the hook ?). Maybe there's something like a 'pre push' hook ?

This is what `pre-commit install` does, it configure it as a git hook to run on every commit.
I personally use it as a pre-push but again, it's up to the developer's preference.

> Another comment is about the role.yaml file as it still looks weird to me. I wonder whether we should careful go through it and drill it down to the stuff that we really need. E.g. we don't need to list Services, or do we ? But that's a question for another PR.

I have the same feeling, but all those things were auto-generated :(

hack/boilerplate.go.txt
```diff
@@ -12,4 +12,4 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
-*/
\ No newline at end of file
+*/
```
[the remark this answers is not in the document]
leseb: No newline at end of file. Agreed it's not very clear.

rhuss:
Thanks @leseb for the elaboration, looks good to me!

For the role issue I've opened https://github.com/llamastack/llama-stack-k8s-operator/issues/18

/lgtm

rhuss (approved):
lgtm

PR #17, opened 2025-05-06
Update go version to 1.23

VaishnaviHire:
This version updagrade is a follow-up to the operator-sdk update where latest versions require  Go 1.23


leseb (approved):
good call thanks

PR #26, opened 2025-05-12
Fix: Kustomize binary installation

VaishnaviHire:
Kustomize to use direct binary download. With latest changes that updates Makefile to use go-install doesn't seem to be working for kustomize. It downloads the binary successfully, but none of the kustomize commands are able to complete execution. Updated to support cross-platform compatibility

- Add platform-specific environment variables (GOOS, GOARCH) using go env
- Enable Go modules with GO111MODULE=on
- Build static binaries with CGO_ENABLED=0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,7 +268,7 @@ set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
 rm -f $(1) || true ;\
-GOBIN=$(LOCALBIN) go install $${package} ;\
+GOBIN=$(LOCALBIN) CGO_ENABLED=0 go install -v $${package} ;\
 mv $(1) $(1)-$(3) ;\
 } ;\
 ln -sf $(1)-$(3) $(1)


Makefile
```diff
@@ -236,7 +236,16 @@ YQ_VERSION ?= v4.45.3
...
 $(KUSTOMIZE): $(LOCALBIN)
-	$(call go-install-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v5,$(KUSTOMIZE_VERSION))
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(KUSTOMIZE)) ;\
+	OS=$$(go env GOOS) && ARCH=$$(go env GOARCH) && \
+	TMPFILE=$$(mktemp) && \
+	curl -sSLo $$TMPFILE https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/$(KUSTOMIZE_VERSION)/kustomize_$(KUSTOMIZE_VERSION)_$${OS}_$${ARCH}.tar.gz && \
+	tar -xzf $$TMPFILE -C $(dir $(KUSTOMIZE)) && \
+	rm $$TMPFILE && \
```
leseb, line 246: Nit: this could be a single command if you pipe the curl to tar. And the TMPFILE won't be needed either.

rhuss:
That's interesting, as I literally copied over the code from `operator-sdk init` --> https://github.com/kubernetes-sigs/kubebuilder/blob/99313dbb243bdfc7c07a9af77f36176237413858/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L269-L272

When this is broken we should add an issue upstream to `kubebuilder`, too.

Makefile
rhuss: Can we use the same technique like `go-install-tool` that creates a `bin/kustomize-<version>` binary and then a symbolic link to `bin/kustomize`. That way it is ensured that you always have the kustomize version running that is mentioned in the Makefile.

rhuss: I mean something like

```
.PHONY: kustomize
kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
$(KUSTOMIZE): $(LOCALBIN)
	@[ -f "$(KUSTOMIZE)-$(KUSTOMIZE_VERSION)" ] || { \
	set -e; \
	echo "Downloading kustomize $(KUSTOMIZE_VERSION)" ;\
	rm -f $(KUSTOMIZE) || true ;\
	OS=$$(go env GOOS) && ARCH=$$(go env GOARCH) ;\
	curl -sSLf https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/$(KUSTOMIZE_VERSION)/kustomize_$(KUSTOMIZE_VERSION)_$${OS}_$${ARCH}.tar.gz | tar xz -C $(dir $(KUSTOMIZE)) ;\
	mv $(dir $(KUSTOMIZE))kustomize $(KUSTOMIZE)-$(KUSTOMIZE_VERSION) ;\
	chmod +x $(KUSTOMIZE)-$(KUSTOMIZE_VERSION) ;\
	} ;\
	ln -sf $(KUSTOMIZE)-$(KUSTOMIZE_VERSION) $(KUSTOMIZE)
```

VaishnaviHire:
> That's interesting, as I literally copied over the code from `operator-sdk init` --> https://github.com/kubernetes-sigs/kubebuilder/blob/99313dbb243bdfc7c07a9af77f36176237413858/pkg/plugins/golang/v4/scaffolds/internal/templates/makefile.go#L269-L272
> 
> When this is broken we should add an issue upstream to `kubebuilder`, too.

I wonder if this is something specific to environment. I am installing this in macOS with 1.23 go version

VaishnaviHire:
@leseb @rhuss It indeed was platform specific issue. Please review my latest changes.

Makefile
```diff
@@ -268,7 +268,12 @@ set -e; \
 package=$(2)@$(3) ;\
 echo "Downloading $${package}" ;\
 rm -f $(1) || true ;\
-GOBIN=$(LOCALBIN) go install $${package} ;\
+GOBIN=$(LOCALBIN) \
+GOOS=$$(go env GOOS) \
+GOARCH=$$(go env GOARCH) \
+GO111MODULE=on \
```
rhuss, line 274: I don't think GOOS and GOARCH are needed, not sure about GO111MODULE. Maybe CGO_ENABLED=0 is good enough.

But if it is needed for sure, happy to merge it.

rhuss (approved):
Thanks!

leseb (approved):

PR #27, opened 2025-05-13
Introduce Operator installation script

VaishnaviHire:
This commit adds changes to improve operator installation

- Users to be able to install operator without cloning the repo
- Readme updated with installation steps

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,15 +49,11 @@ endif
 # Set the Operator SDK version to use. By default, what is installed on the system is used.
 # This is useful for CI or a project to utilize a specific version of the operator-sdk toolkit.
 OPERATOR_SDK_VERSION ?= v1.39.2
-# Image URL to use all building/pushing image targets
-IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.31.0
 
 # Image URL to use all building/pushing image targets
-IMG ?= quay.io/opendatahub/llama-stack-k8s-operator:latest
-# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.24.2
+IMG ?= quay.io/llamastack/llama-stack-k8s-operator:latest
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -49,13 +49,11 @@ endif
...
-# Image URL to use all building/pushing image targets
-IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.31.0
 
 # Image URL to use all building/pushing image targets
-IMG ?= quay.io/opendatahub/llama-stack-k8s-operator:latest
+IMG ?= quay.io/llamastack/llama-stack-k8s-operator:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.24.2
```
rhuss, line 58: Please remove this var, too. My bad, this slipped in during a conflict resolution when rebasing.

Makefile
```diff
@@ -181,9 +179,9 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 
 .PHONY: build-installer
 build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
-	mkdir -p dist
+	mkdir -p release
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default > dist/install.yaml
+	$(KUSTOMIZE) build config/default > release/operator.yaml
```
rhuss, line 184: nit: happy to rename, but if not necessary, my experience is, changing from the operator-sdk templates makes it harder to update it later on for future operator-sdk versions. As long as it doesn't harm and we have no good reasons, we should not deviate from the defaults (even when we don't like it).

rhuss (changes requested):
two minor comments (one is to cleanup my mess :), but overall  looks good to me. Thanks for taking care about the docs!

Makefile
VaishnaviHire: I think this should be okay to point to default image if we want to just do a simple deploy using `make deploy` without any custom image

Makefile
VaishnaviHire: Yeah I can revert this, updated because the naming `dist` did not make sense to me

README.md
```diff
@@ -12,58 +12,33 @@ This repo hosts a kubernetes operator that is responsible for creating and manag
 
 ## Table of Contents
 
-- [Prerequisites](#prerequisites)
-- [Installation](#installation)
 - [Quick Start](#quick-start)
+    - [Installation](#installation)
+    - [Deploying Llama Stack Server](#deploying-the-llama-stack-server)
 - [Developer Guide](#developer-guide)
+    - [Prerequisites](#prerequisites)
```
liangwen12year, line 19: Thanks for making the `README.md` well structured.

Makefile
rhuss: Sorry, I meant `ENVTEST_K8S_VERSION` as it appears twice in the Makefile with different values.

Makefile
rhuss: I see your point and leave it up to you. Not a big issue anyway. Maybe add a small comment that this deviates from the operator-sdk default for a reason, so that during a future alignment of the makefile people don't accidentally revert this change.

Makefile
VaishnaviHire: Updated @rhuss

rhuss (approved):
Thanks! lgtm

PR #30, opened 2025-05-14
ci: introduce mergify bot

leseb:
This enables the Mergify bot to:

* automatically merge PRs under certain conditions using the squash method (like we do today)
* ping autors and label PR when the code has conflicts, making it easy to identify by maintainers too

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,51 @@
+pull_request_rules:
+- name: auto-merge
+  description: >
+    Automatic merge of PRs to main
+  conditions:
+    - "#approved-reviews-by>=1"
+    - "#review-requested=0"
+    - "#changes-requested-reviews-by=0"
+    - base=main
+    - label!=do-not-merge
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -0,0 +1,51 @@
+pull_request_rules:
+- name: auto-merge
+  description: >
+    Automatic merge of PRs to main
+  conditions:
+    - "#approved-reviews-by>=2"
```
rhuss, line 6: since we are effectively only 3 active developers right now, this might become a bottleneck if one is inactive. I would be fine with a single approver, or, if this is possible, with all assigned reviewers approving (so that people might want to ask also for 2 reviewers because they are unsure). 

Or if we still allow manual merge, then having a default of >=2 is fine, too.

leseb: We can still merge manually, but let's relax to 1 for now and put it back to 2 once we have more active maitainers :)

PR #34, opened 2025-05-16
chore: Ensure installer script is updated with manifests

VaishnaviHire:
Updated pre-commit hook to update Installer script on manifest generation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,6 +46,15 @@ repos:
 
 -   repo: local
     hooks:
+      - id: build-installer
+        name: Update Install Script
+        entry: make build-installer
+        language: system
+        pass_filenames: false
+        require_serial: true
+
+-   repo: local
+    hooks:
       - id: generate-api-docs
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -39,7 +39,7 @@ repos:
     hooks:
       - id: generate-manifests
         name: Generate Manifests
-        entry: make generate manifests
+        entry: make generate manifests build-installer
```
leseb, line 42: For troubleshooting purposes, I believe it would be clearer to have separate hooks for each case. This way, it’s easier to identify which make command failed.

.pre-commit-config.yaml
```diff
@@ -44,6 +44,15 @@ repos:
         pass_filenames: false
         require_serial: true
 
+-   repo: local
+    hooks:
+      - id: build-installer
```
leseb, line 49: Nit: since it's a local hook, you can put it inside the previous one. We should move all the local hooks under a single "repo: local". Followup PR

PR #36, opened 2025-05-20
fix: remove trailing line

leseb:
The previous check wasn't entirely correct, since somone can run macOS but use gnu binaries.
Let's remove all ending trailing lines, if any, it doesn't hurt :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -303,7 +303,8 @@ api-docs: crd-ref-docs ## Creates API docs using https://github.com/elastic/crd-
 	@# Combined command to remove .io links, ensure a trailing newline, and collapse multiple blank lines.
 	@sed -i.bak -e '/\.io\/[^v][^1].*)/d' -e '/^$$/N;/^\n$$/D' $(API_DOCS_PATH)
 	@# BSD sed doesn't generate trailing newlines, so no need to remove them.
-	@if [ "$(shell uname)" != "Darwin" ]; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -302,10 +302,8 @@ api-docs: crd-ref-docs ## Creates API docs using https://github.com/elastic/crd-
 	$(CRD_REF_DOCS) --source-path ./ --output-path $(API_DOCS_PATH) --renderer markdown --config ./crd-ref-docs.config.yaml
 	@# Combined command to remove .io links, ensure a trailing newline, and collapse multiple blank lines.
 	@sed -i.bak -e '/\.io\/[^v][^1].*)/d' -e '/^$$/N;/^\n$$/D' $(API_DOCS_PATH)
-	@# BSD sed doesn't generate trailing newlines, so no need to remove them.
-	@if [ "$(shell uname)" != "Darwin" ]; then \
-		sed -i.bak -e '$${/^$$/d}' -e '$${N;/^\n$$/d}' $(API_DOCS_PATH); \
-	fi
+	@# Remove trailing newlines if any.
+	@sed -i.bak -e '$${/^$$/d}' -e '$${N;/^\n$$/d}' $(API_DOCS_PATH); \
```
VaishnaviHire, line 306: This will return error though. The syntax is incompatible with macOS

leseb: Oh good catch.

VaishnaviHire (approved):
/lgtm

PR #37, opened 2025-05-21
chore: Update image pull-policy to Always

VaishnaviHire:
Image pull-policy defaults to `Always` only for :latest tag. Ensure the image is pulled from registry for every updated


cooktheryan (approved):
This will be super helpful as we nail down the distribution we want to ship as the version of llama stack may not change

PR #40, opened 2025-05-29
Factor out Kubernetes resource constructors into reusable builder functions

mfleader:
This PR extracts the inline object‐construction logic from each reconcileXxx (PVC, Deployment, Service, NetworkPolicy) into dedicated Build* helper functions. By centralizing the naming, defaulting, labels, and spec population in pure “builder” routines. It lays the groundwork for an upcoming PR that will wire these builders into an in-memory YAML overlay pipeline (serializing Go objects → filesys → Kustomize → SSA apply).


mfleader:
Decided to go the other route using YAML files to define our runtime resources to facilitate parameter injection outside of the Go code.

PR #41, opened 2025-05-30
Fix Kustomize Deprecation Warnings and Missing Files

mfleader:
Deprecated key usage in our Kustomize files.
```shell
# Warning: 'bases' is deprecated. Please use 'resources' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'commonLabels' is deprecated. Please use 'labels' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'vars' is deprecated. Please use 'replacements' instead. [EXPERIMENTAL] Run 'kustomize edit fix' to update your Kustomization automatically.
```

- [X] `kustomize edit fix --vars` to have Kustomize programmatically fix the issues with the deprecated keys, and the `vars` key (I believe that was the only experimental fix)
- [X] `operator-sdk generate kustomize manifests --interactive=false -q` to generate `config/manifests/bases` and its `llama-stack-k8s-operator.clusterserviceversion.yaml`

### DIFF
--- a/config/manifests/bases/llama-stack-k8s-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/llama-stack-k8s-operator.clusterserviceversion.yaml
@@ -1,0 +1,47 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: '[]'
+    capabilities: Basic Install
+  name: llama-stack-k8s-operator.v0.0.0
+  namespace: placeholder
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+    owned:
+    - displayName: Llama Stack Distribution
+      kind: LlamaStackDistribution
+      name: llamastackdistributions.llamastack.io
+      version: v1alpha1
+  description: Llama Stack K8s Operator description. TODO.
+  displayName: Llama Stack K8s Operator
+  icon:
+  - base64data: ""
+    mediatype: ""
+  install:
+    spec:
+      deployments: null
+    strategy: ""
+  installModes:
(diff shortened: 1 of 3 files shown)


config/manifests/bases/llama-stack-k8s-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: '[]'
+    capabilities: Basic Install
+  name: llama-stack-k8s-operator.v0.0.0
```
VaishnaviHire, line 7: Why do we need to add a CSV here?

mfleader: removed

config/crd/kustomization.yaml
```diff
@@ -28,3 +28,5 @@ resources:
 # the following config is for teaching kustomize how to do kustomization for CRDs.
 configurations:
 - kustomizeconfig.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
rhuss, line 32: I know it doesn't really matter but can we put the apiVersion and kind to the top of the file, as it is the convention for K8s resources (which customize kind of follow) ?

rhuss (changes requested):
actually it looks good for me, but for cosmetic reason I would always move up apiVersion and kind to the top of the file. I know, that the automatic kustomize conversion does not really care, but well :)

VaishnaviHire (approved):

PR #51, opened 2025-06-10
chore: Update mergify checks

VaishnaviHire:


.github/mergify.yml
```diff
@@ -12,6 +12,8 @@ pull_request_rules:
     - check-success=pre-commit
     - check-success=e2e-tests
     - check-success=DCO
+    - check-success=build-latest-image
```
leseb, line 15: this will block PRs on main since the job only runs after the merge

leseb (changes requested):

PR #52, opened 2025-06-10
fix contributing docs on error messages

mfleader:
Missed this [nit](https://github.com/llamastack/llama-stack-k8s-operator/pull/46#discussion_r2138536355) on the previous pull request to add the pre-commit hook to check golang wrapped errors.


leseb (approved):
thanks for the followup

PR #54, opened 2025-06-10
feat: inject operator's runtime pvc definition

mfleader:
- [X] Integrates the Kustomize manifest pipeline with the controller to enable injection of a PVC resource definition.
- [X] Adds the now requisite PVC resource definition file.


pkg/deploy/kustomize.go
```diff
@@ -38,6 +41,16 @@ func ApplyKustomizeManifests(
 	// Use server-side apply for each object so the API server
 	// manages field ownership and merging of concurrent updates.
 	for _, u := range objs {
+		// Set namespace to match the owner. This is crucial for testability
+		// and for the controller to work with namespaced resources.
+		u.SetNamespace(owner.GetNamespace())
+
```
VaishnaviHire, line 47: We could add namespace and label plugins of kustomize here

controllers/llamastackdistribution_controller.go
```diff
@@ -152,45 +155,81 @@ func (r *LlamaStackDistributionReconciler) SetupWithManager(mgr ctrl.Manager) er
...
-		return fmt.Errorf("failed to fetch PVC: %w", err)
+		return fmt.Errorf("failed to read base kustomization.yaml: %w", err)
 	}
-	// PVCs are immutable after creation, so we don't need to update them
+	pvcFile, err := r.Fs.ReadFile(filepath.Join(baseDir, "pvc.yaml"))
+	if err != nil {
+		return fmt.Errorf("failed to read base pvc.yaml: %w", err)
+	}
+
+	opFs.MkdirAll(baseDir)
```
VaishnaviHire, line 186: Why do we need this, the [function](https://github.com/llamastack/llama-stack-k8s-operator/pull/54/files#diff-d2fef16128effe03ba3c644a0ead9cdbc473e2a47a8266c9c2f20960798325b1L62) takes manifest directory as input

controllers/llamastackdistribution_controller.go
```diff
@@ -152,45 +155,81 @@ func (r *LlamaStackDistributionReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+	opFs.WriteFile(filepath.Join(baseDir, "kustomization.yaml"), kustFile)
+	opFs.WriteFile(filepath.Join(baseDir, "pvc.yaml"), pvcFile)
+
+	// Create the overlay directory in the in-memory filesystem.
+	overlayDir := "overlay"
```
VaishnaviHire, line 191: Why is this needed?

VaishnaviHire: Discussed with @mfleader, using [kustomize vars](https://github.com/openshift/kubernetes-kubectl/blob/master/docs/book/pages/reference/kustomize.md#vars) should resolve this.

Dockerfile
```diff
@@ -23,6 +24,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
+COPY --from=builder /workspace/manifests/ ./manifests/
```
leseb, line 27: Instead of adding the manifests to the container image, why not embed them in the Go binary using [go embed](https://pkg.go.dev/embed). I believe this would simplify some of the FS setup and file load?

leseb: I'm not sure I follow why it's more difficult. The current approach seems to involve more steps than using `embed`, since you need to include the COPY steps in the Dockerfile. 
Also things like `"manifests/base"` are hardcoded so how it different than embed?

Am I missing something?

controllers/llamastackdistribution_controller.go
```diff
@@ -48,16 +50,19 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
+//go:embed all:manifests/base
+var manifests embed.FS
```
leseb, line 54: ❤️

mfleader: lol does this mean the conversation is resolved?

Dockerfile
leseb: yesss

controllers/llamastackdistribution_controller.go
```diff
@@ -48,10 +48,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+<<<<<<< HEAD
```
rhuss, line 51: @mfleader looks like you've checked in some unresolved git conflict.

PR #60, opened 2025-06-13
fix: avoid memory growth with logger instance

leseb:
Each reconcile was re-assigning the values in the the context so the logger was growing exponentially. Now, we instantiate the logger once and re-create it when needed.


controllers/llamastackdistribution_controller.go
```diff
@@ -70,15 +69,17 @@ type LlamaStackDistributionReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.2/pkg/reconcile
 func (r *LlamaStackDistributionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	r.Log = r.Log.WithValues("llamastack", req.NamespacedName)
+	log := log.FromContext(ctx).WithValues("namespace", req.Namespace, "name", req.Name)
```
mfleader, line 72: can we add your brief summary of why we instantiated and configured the logger this way?

mfleader (changes requested):

mfleader (approved):

PR #65, opened 2025-06-16
fix: avoid repeating name and namespace log values

leseb:
The first entry in the reconcile loop already initializes a logger with the name and namespace values so there is no need to re-add those.


VaishnaviHire:
@Mergifyio rebase

VaishnaviHire (approved):
Thanks !

PR #72, opened 2025-06-18
Add support for additional fields

VaishnaviHire:
This commit allows users to include additional fields like annotations, sidecars, labels in server deployment

This change is needed to support adding `otel` sidecar and annotations for observability


VaishnaviHire:
> I think all other Apply* deserve a similar treatment. We can do this in a followup PR.
Yes. I think these will be addressed anyways in follow-ups to #54

PR #75, opened 2025-06-20
feat(field_mutator): add kustomize plugin to support field patching

mfleader:
- [X] Adds a Kustomize plugin to support setting default and arbitrary values for any given field.

### DIFF
--- a/pkg/deploy/plugins/field_mutator.go
+++ b/pkg/deploy/plugins/field_mutator.go
@@ -102,6 +102,13 @@ func setTargetField(res *resource.Resource, value any, mapping FieldMapping) err
 		return fmt.Errorf("failed to unmarshal YAML: %w", unmarshalErr)
 	}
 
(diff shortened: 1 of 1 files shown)


VaishnaviHire (approved):

pkg/deploy/plugins/field_mutator.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	var data map[string]any
+	if unmarshalErr := yaml.Unmarshal(yamlBytes, &data); unmarshalErr != nil {
+		return fmt.Errorf("failed to unmarshal YAML: %w", unmarshalErr)
+	}
+
+	fields := strings.Split(mapping.TargetField, ".")
+	current := data
+	for _, field := range fields[:len(fields)-1] {
```
leseb, line 107: This probably deserves a comment on why we exclude the last one?

leseb (commented):
nit otherwise LGTM!

PR #78, opened 2025-06-24
(feat): Improvements in Dockerfile

VaishnaviHire:
- Update base image to UBI9 minimal for FIPS compatibility
- Maintain multi-architecture build support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
 # Build the manager binary
-FROM golang:1.23 as builder
+ARG GOLANG_VERSION=1.23
+
+FROM registry.access.redhat.com/ubi8/go-toolset:${GOLANG_VERSION} as builder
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,10 @@
 # Build the manager binary
-FROM golang:1.23 as builder
+ARG GOLANG_VERSION=1.23
+
+FROM registry.access.redhat.com/ubi8/go-toolset:${GOLANG_VERSION} as builder
+ARG TARGETOS
+ARG TARGETARCH
+ARG CGO_ENABLED
```
leseb, line 7: Should we stay consistent with `GOLANG_VERSION`and do:

```
ARG TARGETOS=linux
ARG TARGETARCH
ARG CGO_ENABLED=0
``` 

Instead of the bash expansion below?

PR #81, opened 2025-06-25
feat(set_ns): add kustomize plugin to set a resource's namespace

mfleader:
- [X] Adds a Kustomize plugin to set a given Resource's Namespace to the Operator owner instance's Namespace.
- [X] Factors out `Name` validation behavior into `plugins.go` for reuse in `namespace.go`.

### DIFF
--- a/pkg/deploy/plugins/namespace_test.go
+++ b/pkg/deploy/plugins/namespace_test.go
@@ -77,7 +77,7 @@ func TestNamespacePlugin(t *testing.T) {
 		assert.Equal(t, testNamespace, transformedSvc.GetNamespace())
 	})
 
-	t.Run("fails with empty eamespace", func(t *testing.T) {
+	t.Run("fails with empty namespace", func(t *testing.T) {
 		resMap := resmap.New()
 		// create any namespaced resource
 		dep := newTestResource(t, "apps/v1", "Deployment", "my-app", "", nil)
@@ -85,7 +85,8 @@ func TestNamespacePlugin(t *testing.T) {
 
 		_, err := CreateNamespacePlugin("")
 		require.Error(t, err)
-		assert.Contains(t, err.Error(), "namespace cannot be empty")
+		assert.Contains(t, err.Error(), "namespace")
+		assert.Contains(t, err.Error(), "empty")
 	})
 
 	t.Run("overwrites namespace", func(t *testing.T) {


pkg/deploy/plugins/namespace.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Config implements the TransformerPlugin interface.
+func (p *namespacePlugin) Config(h *resmap.PluginHelpers, config []byte) error {
+	return nil
+}
+
+// Transform implements the TransformerPlugin interface.
+func (p *namespacePlugin) Transform(m resmap.ResMap) error {
+	// do not transform an invalid namespace
+	if p.namespace == "" {
```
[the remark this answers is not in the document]
mfleader: I think it should be since the k8s api will tell us if it is a valid Namespace name before it creates the Namespace, but I also just added this exact validation check in the [name_prefix pr](https://github.com/llamastack/llama-stack-k8s-operator/pull/77) in that plugin's [Transform()](https://github.com/mfleader/llama-stack-k8s-operator/blob/97a94c1c46833dac925965ca269b65767c32f4cf/pkg/deploy/plugins/name_prefix.go#L53), so it's an easy change to factor that out into a `plugin.go` file of shared behavior.

pkg/deploy/plugins/namespace.go
```diff
@@ -0,0 +1,42 @@
...
+		return errors.New("failed to set namespace: namespace cannot be empty")
+	}
+	for _, res := range m.Resources() {
+		// skip cluster-scoped resources because they don't have a namespace
+		if res.GetGvk().IsClusterScoped() {
+			continue
+		}
+		if err := res.SetNamespace(p.namespace); err != nil {
```
rhuss, line 37: So if I understand right, it will override the namespace unconditionally even when its already set on the resource, and not only when the namespace is missing ?

rhuss:
@mfleader @leseb @VaishnaviHire @skamenan7 @rhdedgar @elbehery

While working on the PR (pardonnez-mois for hijacking the issue :), I would quickly raise a proposal:

* Adding a `lgtm-with-suggestions` that a PR reviewer can add when she only has optional suggestions to add, but no hard feelings and don't want to prevent merging the PR. It's just to seed some proposal that are not hard arguments (stuff, that we usually name as `nits`)
* The author of the PR the can remove that label to trigger an auto-merge (when the other conditions are met, too) or add the suggestions (in this case, a new approval is needed, and hopefully the automation could remove that label on additional change, much like approvals are removed after additional changes)

This allows us to have some more flexibility, moves back some responsibility to the author and hopefully helps in unblocking and speeding-up things.

wdyt ?

pkg/deploy/plugins/namespace.go
mfleader: @VaishnaviHire what do you think? I'm okay with the namespace plugin as a forced write, but is that intended behavior?

mfleader:
I agree with @leseb on their points, except I did not think `nits` had to be addressed, but I'm also happy to accept their definition. Additionally, I think it is time for this conversation to be moved from this pr b/c it is now blocking the conclusion and merging of this pr.

pkg/deploy/plugins/namespace.go
VaishnaviHire: @rhuss Since these are internal manifests, managed by the operator explicitly, we do want to override any `namespace` value that doesn't correspond to the given namespace. We are creating these runtime plugins to override the defaults in the manifests.

Issue #53, opened 2025-06-10
GitHub Actions should use SHA-pinned actions


leseb:
Only a few use a SHA1, they all should so that dependabot can offer updates:

```
 grep -RE '^.*uses:[^@]+@[^ ]+' .github/workflows/                    
.github/workflows/pre-commit.yml:        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
.github/workflows/pre-commit.yml:        uses: actions/setup-go@v5
.github/workflows/pre-commit.yml:      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
.github/workflows/build-image.yml:        uses: actions/setup-go@v3
.github/workflows/build-image.yml:        uses: actions/checkout@v3
.github/workflows/build-image.yml:        uses: docker/login-action@v2
.github/workflows/run-e2e-test.yml:        uses: actions/checkout@v4
.github/workflows/run-e2e-test.yml:        uses: actions/setup-go@v5
.github/workflows/run-e2e-test.yml:        uses: docker/setup-buildx-action@v3
.github/workflows/run-e2e-test.yml:        uses: helm/kind-action@v1
.github/workflows/code-coverage.yml:        uses: actions/checkout@v4
.github/workflows/code-coverage.yml:        uses: actions/setup-go@v5
.github/workflows/code-coverage.yml:        uses: GoTestTools/limgo-action@v1.0.2
.github/workflows/code-coverage.yml:        uses: actions/upload-artifact@v4
``` 

Success criterias:

* use SHA1 for all `uses` statements
* add a precommit check - we can use the one we wrote in https://github.com/meta-llama/llama-stack/pull/2159

Evergreenies:
Raised a PR #84  with suggested changes.

PR #85, opened 2025-06-27
[ACTION] Applied SHA-pinned actions #53

Evergreenies:
Applied SHA as per the issue #53 

- [x] Applied SHA1 for all actions (`uses` statements)
- [x] Added a pre-commit check
- [x] Fixed *.yaml linting/formatting

```
 grep -RE '^.*uses:[^@]+@[^ ]+' .github/workflows/
.github/workflows/release-image.yml:        uses: actions/setup-go@be3c94b385c4f180051c996d336f57a34c397495 # v3.6.1
.github/workflows/release-image.yml:        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
.github/workflows/release-image.yml:        uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
.github/workflows/run-e2e-test.yml:        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
.github/workflows/run-e2e-test.yml:        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
.github/workflows/run-e2e-test.yml:        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
.github/workflows/run-e2e-test.yml:        uses: helm/kind-action@a1b0e391336a6ee6713a0583f8c6240d70863de3 # v1.12.0
.github/workflows/run-e2e-test.yml:        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
.github/workflows/pre-commit.yml:        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
.github/workflows/pre-commit.yml:        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
.github/workflows/pre-commit.yml:      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
.github/workflows/build-image.yml:        uses: actions/setup-go@be3c94b385c4f180051c996d336f57a34c397495 # v3.6.1
.github/workflows/build-image.yml:        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
.github/workflows/build-image.yml:        uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
.github/workflows/code-coverage.yml:        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
.github/workflows/code-coverage.yml:        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
.github/workflows/code-coverage.yml:        uses: GoTestTools/limgo-action@d4a725a46ab3bdbbf23186bbbe54b4ff4a35d5a7 # v1.0.2
.github/workflows/code-coverage.yml:        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```


Evergreenies:
> do we really want the auto-fix by calling `args: [pin, --inplace]` ?
> 
> I can see one use case when a new GH action is to be added in the commit, committer would like to take advantage of getting it updated before making the commit. Otherwise, it should not get called in each pre-commit run.

Right now yes it is auto fix and it fixes only when there is no SHA found for the action.
There are two possibilities may be,
- Make use of `args: [pin, --diff]` to show differences with SHA while pre-commit.
- Or, we have updated SHA, so we can remove that auto fix step from pre-commit.
You can suggest.

zdtsw:
> > do we really want the auto-fix by calling `args: [pin, --inplace]` ?
> > I can see one use case when a new GH action is to be added in the commit, committer would like to take advantage of getting it updated before making the commit. Otherwise, it should not get called in each pre-commit run.
> 
> Right now yes it is auto fix and it fixes only when there is no SHA found for the action. There are two possibilities may be,
> 
> * Make use of `args: [pin, --diff]` to show differences with SHA while pre-commit.
> * Or, we have updated SHA, so we can remove that auto fix step from pre-commit.
>   You can suggest.

I would guess, with your current PR which has all versions updated with SHA, this should have covered original "issue" + `hack/check-workflows-uses-hashesh.sh` is enough to warn such in the future.

Personally  I do not see the rest [changes](https://github.com/llamastack/llama-stack-k8s-operator/blob/9e9e50d75a7588f89f6c275b9ef096be9c586065/.pre-commit-config.yaml#L29-L40) can do harm, but i defer this to @leseb 

btw, should https://github.com/llamastack/llama-stack-k8s-operator/blob/9e9e50d75a7588f89f6c275b9ef096be9c586065/.pre-commit-config.yaml#L8 be updated as well ?

Evergreenies:
> > > do we really want the auto-fix by calling `args: [pin, --inplace]` ?
> > > I can see one use case when a new GH action is to be added in the commit, committer would like to take advantage of getting it updated before making the commit. Otherwise, it should not get called in each pre-commit run.
> > 
> > 
> > Right now yes it is auto fix and it fixes only when there is no SHA found for the action. There are two possibilities may be,
> > 
> > * Make use of `args: [pin, --diff]` to show differences with SHA while pre-commit.
> > * Or, we have updated SHA, so we can remove that auto fix step from pre-commit.
> >   You can suggest.
> 
> I would guess, with your current PR which has all versions updated with SHA, this should have covered original "issue" + `hack/check-workflows-uses-hashesh.sh` is enough to warn such in the future.
> 
> Personally I do not see the rest [changes](https://github.com/llamastack/llama-stack-k8s-operator/blob/9e9e50d75a7588f89f6c275b9ef096be9c586065/.pre-commit-config.yaml#L29-L40) can do harm, but i defer this to @leseb
> 
> btw, should
> 
> https://github.com/llamastack/llama-stack-k8s-operator/blob/9e9e50d75a7588f89f6c275b9ef096be9c586065/.pre-commit-config.yaml#L8
> 
> be updated as well ?

Thank you for pointing that out, I have updated that to fixed SHA. 
And, as our original issue is covered by updating SHA I have removed auto-fix action. If we need to integrate that then will check that out in future.

rhdedgar (approved):
I think that covers everything. Thank you!

leseb (approved):

PR #89, opened 2025-07-01
feat(kustomizer): enable kustomize plugin support

mfleader:
- [X] Adds support for custom plugins, so there is no regression in the controller's ability to prefix resource names, and set the namespace on resources.
- [X] Refactor Kustomize component to separate concerns between rendering Kustomize resources and Server Side Applying those Kustomize resources.


pkg/deploy/kustomizer.go
```diff
@@ -0,0 +1,209 @@
...
+	// overwriting a resource that was created by another user or controller.
+	isOwner := false
+	for _, ref := range existing.GetOwnerReferences() {
+		if ref.UID == ownerInstance.GetUID() {
+			isOwner = true
+			break
+		}
+	}
+	if !isOwner {
+		return fmt.Errorf("failed to patch resource %s/%s because it is not owned by this instance",
```
VaishnaviHire, line 149: If we don't own the resource, we should not return error. Just need to return nil. We can just log this error.

rhdedgar (approved):
Looks good to me, thanks!

PR #95, opened 2025-07-03
feat: use embedded distributions data instead of ConfigMap

VaishnaviHire:
Replace ConfigMap-based distribution management with go:embed approach to prevent upgrade failures and ensure version consistency.


leseb (commented):
What are the benefits of this method as opposed to embedding the list in our codebase?

The file is just used in the codebase, not outside (like the yaml used to be)... :)

VaishnaviHire:
> What are the benefits of this method as opposed to embedding the list in our codebase?
> 
> The file is just used in the codebase, not outside (like the yaml used to be)... :)

The only advantage I see is this will help us maintain different distribution files depending on installations(midstream, upstream etc)

leseb:
> > What are the benefits of this method as opposed to embedding the list in our codebase?
> > The file is just used in the codebase, not outside (like the yaml used to be)... :)
> 
> The only advantage I see is this will help us maintain different distribution files depending on installations(midstream, upstream etc)

But we could do that in code too right?

leseb (commented):
Editing a regular file is easier than editing go code. Can we just make this file is validated a valid json so we don't break the serialization? pre-commit has this https://github.com/llamastack/llama-stack-k8s-operator/blob/main/.pre-commit-config.yaml#L24 so we might be good but could you validate? Thanks!

leseb (approved):

PR #97, opened 2025-07-07
refactor: consolidate security annotations

### LINKED ISSUE #18: Refactor: Consolidate security annotations
The annotations in https://github.com/llamastack/llama-stack-k8s-operator/blob/main/controllers/kubebuilder_rbac.go should be revisited and ...

* ... avoid `*` matches and list the required verbs directly
* ... verify if all access is really needed. It looks a little bit too open to me
* ... combine resources where possible to make the resulting roles.yaml file as compact as possible.

rhdedgar:
Trimming down existing wildcard RBAC permissions via kubebuilder markers.

Tested on an OpenShift 4.18.17 | Kubernetes v1.31.9 cluster.

Closes #18

### DIFF
--- a/controllers/kubebuilder_rbac.go
+++ b/controllers/kubebuilder_rbac.go
@@ -1,26 +1,21 @@
 package controllers
 
+// LlamaStackDistribution CRD permissions
 //+kubebuilder:rbac:groups=llamastack.io,resources=llamastackdistributions,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=llamastack.io,resources=llamastackdistributions/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=llamastack.io,resources=llamastackdistributions/finalizers,verbs=update
 
-// +kubebuilder:rbac:groups="core",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
+// Deployment permissions - controller creates and manages deployments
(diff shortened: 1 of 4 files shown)


controllers/kubebuilder_rbac.go
```diff
@@ -1,26 +1,21 @@
...
-// +kubebuilder:rbac:groups="core",resources=persistentvolumes,verbs=*
-// +kubebuilder:rbac:groups="core",resources=persistentvolumeclaims,verbs=*
+// PVC permissions - controller creates PVCs (immutable after creation, no update/patch needed)
+//+kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=get;list;watch;create
 
-// +kubebuilder:rbac:groups="apps",resources=deployments/finalizers,verbs=*
-// +kubebuilder:rbac:groups="core",resources=deployments,verbs=*
-// +kubebuilder:rbac:groups="apps",resources=deployments,verbs=*
+// ConfigMap permissions - controller reads user configmaps and creates operator config configmaps
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create
```
zdtsw, line 18: wont require "update" if https://github.com/llamastack/llama-stack-k8s-operator/blob/main/config/manager/distribution-configmap.yaml changes in new release?

rhdedgar: That's true! It would update instead of delete & re-create in that case.

I've updated the `configmaps` verbs to include `update` and `patch`.

VaishnaviHire:
@Mergify rebase

PR #98, opened 2025-07-11
chore: Add LICENSE

VaishnaviHire:
Add Apache License to the llama-stack operator


raghotham (changes requested):
Are you ok with MIT license which is what we are using for the other repos?

rhuss (approved):
Lgtm, thx!

raghotham (approved):

PR #109, opened 2025-07-16
feat: support providers with self-signed certificates

rhdedgar:
- Updated the operator's CRD to allow the definition of a configMap containing ca-cert data, by configuring fields in a LlamaStackDistribution CR.
- Configured llama-stack to trust connections to providers like vLLM using server certificates from that Certificate Authority.
- Added an example script to generate generic self-signed certificates in PEM format to use with vLLM.
- Added an example configuration file to load the self-signed certs into a secret, mount the secret into the container, and point vLLM to the certs on server startup.


rhdedgar:
Current blocker:
- The e2e tests work on an OpenShift cluster with sufficient resources, but hang when running in github's current runner env for this repo.

The e2e test may have to be marked as optional, unless more resources can be allocated to the test runner,

docs/ca-bundle-configuration.md
```diff
@@ -0,0 +1,282 @@
+# CA Bundle Configuration for LlamaStackDistribution
```
VaishnaviHire, line 1: We can just add a section in README with a reference to updated spec, similar to ConfigMap [changes](https://github.com/llamastack/llama-stack-k8s-operator?tab=readme-ov-file#using-a-configmap-for-runyaml-configuration). 

These docs(ca-bundle-configuration.md & vllm-tls-testing.md), although helpful seem to be more of helper documents. Maybe we can move them under `docs/additional` or `docs/guides` ?

cdoern (commented):
combed through the code , it all looks good! One question about error handling

rhdedgar:
Being able to narrow down the tests to just the operator-managed portions helps a lot. Trimming the vLLM setup and connection piece allowed for a huge reduction in complexity, and the GH tests are currently passing.

I'll look into the comments around [docs, etc] now.

config/samples/generate_certificates.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rhdedgar: Do you mean remove the entire file, or just that line? At the moment, it generates an example self-signed CA that gets added to the ConfigMap, which gets used by the LlamaStackDistribution pod. It mounts the ConfigMap into the llama stack server container during the test.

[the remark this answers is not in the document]
VaishnaviHire: You can ignore this comment, just saw the e2e test references

config/crd/bases/llamastack.io_llamastackdistributions.yaml
```diff
@@ -2001,6 +2001,31 @@ spec:
...
+                        properties:
+                          configMapName:
+                            description: ConfigMapName is the name of the ConfigMap
+                              containing CA bundle certificates
+                            type: string
+                          configMapNamespace:
+                            description: ConfigMapNamespace is the namespace of the
+                              ConfigMap (defaults to the same namespace as the CR)
+                            type: string
+                          key:
```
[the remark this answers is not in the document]
rhdedgar: Good stuff, I've updated key->configMapKey for tlsConfig.caBundle.

controllers/resource_helper.go
```diff
@@ -77,20 +113,48 @@ func buildContainerSpec(instance *llamav1alpha1.LlamaStackDistribution, image st
...
+	// Add CA bundle volume mount if TLS config is specified
+	if instance.Spec.Server.TLSConfig != nil && instance.Spec.Server.TLSConfig.CABundle != nil {
+		caBundleKey := instance.Spec.Server.TLSConfig.CABundle.ConfigMapKey
+		if caBundleKey == "" {
+			caBundleKey = "ca-bundle.crt"
+		}
+
+		container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{
+			Name:      "ca-bundle",
+			MountPath: "/etc/ssl/certs/" + caBundleKey,
```
leseb, line 127: The path should be determined based on the Linux distro. Ideally, we would put this in `/etc/pki/ca-trust/source/anchors/` and then run `sudo update-ca-trust` so we don't even need to pass the cert later to the vLLM client. I might be stretching so we can revisit later, any thoughts?

leseb (commented):
The current code looks good to me, I've left some comments but there is no action required. We can follow up and take care of the code hygiene later.

However, I'd like to be sure we are good on the cert path so we don't change that later.

Thanks!

config/crd/bases/llamastack.io_llamastackdistributions.yaml
```diff
@@ -2001,6 +2001,31 @@ spec:
                         pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                         x-kubernetes-int-or-string: true
                     type: object
+                  tlsConfig:
+                    description: TLSConfig defines the TLS configuration for the llama-stack
+                      server
+                    properties:
+                      caBundle:
```
jgarciao, line 2008: @rhdedgar , RHOAI's operator creates the ConfigMap odh-trusted-ca-bundle, which contains 2 fields:
 - ca-bundle.crt > The Cluster Network Operator (CNO) injects the cluster-wide CA here
 - odh-ca-bundle.crt > RHOAI's operator adds the certificates specified by the user in DSCInitialization (spec.trustedCABundle.customCABundle)

In order to support both fields (more info in my [comment](https://issues.redhat.com/browse/RHAIENG-188?focusedId=27620503&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27620503) at RHAIENG-188), would it make sense to declare caBundles as an array so multiple configmaps and keys could be used?

CC: @VaishnaviHire 
```
 caBundles:
      description: List of CA bundle configurations for custom certificates
      type: array
      items:
        type: object
        properties:
          configMapKey:
            default: ca-bundle.crt
            description: ConfigMapKey is the key within the ConfigMap
              that contains the CA bundle data (defaults to "ca-bundle.crt")
            type: string
          configMapName:
            description: ConfigMapName is the name of the ConfigMap
              containing CA bundle certificates
            type: string
          configMapNamespace:
            description: ConfigMapNamespace is the namespace of the
              ConfigMap (defaults to the same namespace as the CR)
            type: string
        required:
          - configMapName
```

controllers/resource_helper.go
rhdedgar: That would be more convenient, but I didn't find `/etc/pki/ca-trust/source/anchors/` available in the LlamaStackDistribution container (it's a Debian-based container image). So I went with a cross-platform middle ground solution.

Do you mean dynamically detect if `/etc/pki/ca-trust/source/anchors/` exists, and use it only if it is available?

controllers/resource_helper.go
```diff
@@ -61,36 +73,170 @@ func buildContainerSpec(instance *llamav1alpha1.LlamaStackDistribution, image st
...
+func createCABundleInitContainer(caBundleConfig *llamav1alpha1.CABundleConfig) corev1.Container {
+	// Build the shell command to concatenate all specified keys
+	keyPaths := make([]string, 0, len(caBundleConfig.ConfigMapKeys))
+
+	for _, key := range caBundleConfig.ConfigMapKeys {
+		keyPaths = append(keyPaths, "/tmp/ca-source/"+key)
+	}
+
+	// Command to concatenate all keys into the target file
+	command := fmt.Sprintf("cat %s > /etc/ssl/certs/ca-bundle.crt", strings.Join(keyPaths, " "))
```
leseb, line 218: This is dangerous, we are interpolating user-controlled (keyPaths) directly into a shell command string. This risks a shell-injection attack. 

We must do a bit more validation on `keyPaths` first by escaping shell metacharacters. Or use a library like https://github.com/alessio/shellescape

controllers/resource_helper.go
```diff
@@ -99,73 +245,141 @@ func configurePodStorage(instance *llamav1alpha1.LlamaStackDistribution, contain
...
-			},
-			SecurityContext: &corev1.SecurityContext{
-				RunAsUser:  ptr.To(int64(0)), // Run as root to be able to change ownership
-				RunAsGroup: ptr.To(int64(0)),
-			},
-		}
+	commands := []string{
+		fmt.Sprintf("mkdir -p %s", mountPath),
+		fmt.Sprintf("(chown 1001:0 %s 2>/dev/null || echo 'Warning: Could not change ownership')", mountPath),
+		fmt.Sprintf("ls -la %s", mountPath),
```
leseb, line 293: This could fail too on permission denied. Pipe it too?

controllers/resource_helper.go
```diff
@@ -99,73 +245,141 @@ func configurePodStorage(instance *llamav1alpha1.LlamaStackDistribution, contain
...
-					MountPath: mountPath,
-				},
-			},
-			SecurityContext: &corev1.SecurityContext{
-				RunAsUser:  ptr.To(int64(0)), // Run as root to be able to change ownership
-				RunAsGroup: ptr.To(int64(0)),
-			},
-		}
+	commands := []string{
+		fmt.Sprintf("mkdir -p %s", mountPath),
```
leseb, line 291: This could fail too on permission denied. Pipe it too?

nathan-weinberg (commented):
Bit outside my wheelhouse but overall nothing jumped out to me as blocking, just one minor comment

rhdedgar:
I've marked this as a Draft so it's more clear that it's a WIP. A few comments have come in on pieces that I'm not going to keep in the final PR since the last round of scope / date changes.

controllers/resource_helper.go
```diff
@@ -61,111 +109,428 @@ func buildContainerSpec(instance *llamav1alpha1.LlamaStackDistribution, image st
...
-				"-c",
-				fmt.Sprintf("chown --verbose --recursive 1001:0 %s", mountPath),
-			},
-			VolumeMounts: []corev1.VolumeMount{
-				{
-					Name:      "lls-storage",
-					MountPath: mountPath,
-				},
+	// Use a secure script approach that doesn't involve dynamic command construction
+	// This eliminates the possibility of command injection
```
leseb, line 276: 👍🏻

config/crd/bases/llamastack.io_llamastackdistributions.yaml
jgarciao: Thanks!

api/v1alpha1/llamastackdistribution_types.go
```diff
@@ -92,6 +95,38 @@ type UserConfigSpec struct {
...
 
+// TLSConfig defines the TLS configuration for the llama-stack server
+type TLSConfig struct {
+	// CABundle defines the CA bundle configuration for custom certificates
+	// +optional
+	CABundle *CABundleConfig `json:"caBundle,omitempty"`
+}
+
+// CABundleConfig defines the CA bundle configuration for custom certificates
+// +kubebuilder:validation:XValidation:rule="!(has(self.configMapKey) && has(self.configMapKeys))",message="Only one of configMapKey or configMapKeys can be specified"
```
VaishnaviHire, line 106: This validation fails when I add `configMapKeys`

```
Danger alert:An error occurred
Error "Invalid value: "object": Only one of configMapKey or configMapKeys can be specified" for field "spec.server.tlsConfig.caBundle".
```

This is due to deafult value for `ConfigMapKey`. Remove the default -
```
// +kubebuilder:default:="ca-bundle.crt"
```

Alternatively we could just have only one field, `ConfigMapKeys`

controllers/resource_helper.go
```diff
@@ -61,111 +109,428 @@ func buildContainerSpec(instance *llamav1alpha1.LlamaStackDistribution, image st
...
+		container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{
+			Name:      "ca-bundle",
+			MountPath: "/etc/ssl/certs/" + caBundleKey,
+			SubPath:   caBundleKey,
+			ReadOnly:  true,
+		})
+	} else if r != nil {
+		// Check for auto-detected ODH trusted CA bundle
+		if _, keys, err := r.detectODHTrustedCABundle(ctx, instance); err == nil && len(keys) > 0 {
+			// Mount the auto-detected consolidated CA bundle
```
VaishnaviHire, line 212: Shouldn't all the cert keys be consolidated in one single file, `SSL_CERT_FILE`? While testing, I see its not the case

api/v1alpha1/llamastackdistribution_types.go
rhdedgar: +1 to removing ConfigMapKey and just keeping ConfigMapKeys. I've made these changes in a local branch and will push once done addressing the recent comments.

controllers/resource_helper.go
rhdedgar: Interesting, I was able to confirm that it does seem there has been a regression since opening this PR. I haven't tracked down the issue yet, I'll continue troubleshooting and see what's causing it.

rhdedgar: Ok, the concatenation bug should now be resolved. I've updated the PR with the latest changes from my branch, and the tests are passing.

VaishnaviHire:
@mergify rebase

PR #111, opened 2025-07-17
Update PVC permissions for LLS

VaishnaviHire:
This commit fixes the write permissions for PVC 
### Testing

Updated permissions-
Screenshot 2025-07-17 at 08 41 22

### DIFF
--- a/controllers/resource_helper.go
+++ b/controllers/resource_helper.go
@@ -22,6 +22,7 @@ import (
 
 	llamav1alpha1 "github.com/llamastack/llama-stack-k8s-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 )
 
 // buildContainerSpec creates the container specification.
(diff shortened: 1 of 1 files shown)


controllers/resource_helper.go
```diff
@@ -109,6 +110,34 @@ func configurePodStorage(instance *llamav1alpha1.LlamaStackDistribution, contain
...
+			mountPath = instance.Spec.Server.Storage.MountPath
+		}
+
+		initContainer := corev1.Container{
+			Name:  "update-pvc-permissions",
+			Image: "registry.access.redhat.com/ubi9/ubi-minimal:latest",
+			Command: []string{
+				"/bin/sh",
+				"-c",
+				fmt.Sprintf("chown --verbose --recursive 1001:0 %s", mountPath),
```
leseb, line 126: that assumes user 1001 exists in the distribution image :)

leseb: Now I'm wondering if `chmod` would have been easier, so we don't have to guess the user ID...

PR #116, opened 2025-07-17
(chore) Update manager deployment spec

VaishnaviHire:
- Update memory resource limit to avoid OOMKilled error 
        - As we are watching all configmaps in cluster, we hit this error for larger clusters . We can reconfigure the values once we add optimization for ConfigMap watch.
- Update imagePullPolicy to always, for easier testing

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -42,6 +42,7 @@ spec:
         - name: OPERATOR_VERSION
           value: "latest"
         image: controller:latest
+        imagePullPolicy: Always
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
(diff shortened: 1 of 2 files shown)


config/manager/manager.yaml
```diff
@@ -65,9 +66,9 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 128Mi
+            memory: 1Gi
           requests:
             cpu: 10m
-            memory: 64Mi
+            memory: 128Mi
```
leseb, line 72: Should we bump this to 256 based on your comment?

rhdedgar (commented):
+1, with a question - will we ever want to set `imagePullPolicy: IfNotPresent` at some point once the operator is GA?

At the moment `imagePullPolicy: Always` is already used by default for images with the tag `latest`.

PR #118, opened 2025-07-18
refactor(field_mutator): migrate dot notation to JSON Pointer syntax

mfleader:
Replaces custom dot notation parser with [RFC 6901 JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) syntax to handle Kubernetes label keys containing dots (e.g., `app.kubernetes.io/instance`). 

Migration enables proper Service selector transformations where previous dot notation failed on keys like `spec.selector["app.kubernetes.io/instance"]`, unblocking dynamic Service-to-Deployment label matching.


leseb (approved):
nice to get ride of the manual parsing, thanks!

rhdedgar (approved):
Looks good to me!

PR #131, opened 2025-08-01
(fix): Preserve immutable fields during upgrade

VaishnaviHire:
During upgrades, if we do server-side apply without preserving the selectors field, we get 'field is immutable' error which blocks deployment updates and also causes continuous operator reconciliation. Preserving exists selector values fixes the issue.


mfleader (changes requested):
Do you have time to add a regression test since our test suite didn't catch this? If not I can write one as a follow up.

PR #133, opened 2025-08-04
feat(controllers): enable testing of provider and version status via dependency injection

mfleader:
Refactors the LlamaStackDistributionReconciler to accept an httpClient
dependency. This makes the controller's status-update logic, which relies
on external API calls, testable in isolation.

The primary goal of this change is to enable a new integration test that
validates the controller correctly populates provider and version info
in the resource status. This test is added to formalize the expected
behavior and prevent regressions.

`checkHealth()` and its endpoint `v1/health` were not included in these changes since they will be removed in a subsequent code change, [130](https://github.com/llamastack/llama-stack-k8s-operator/pull/130).

### DIFF
--- a/controllers/llamastackdistribution_controller_test.go
+++ b/controllers/llamastackdistribution_controller_test.go
@@ -384,7 +384,11 @@ func TestLlamaStackProviderAndVersionInfo(t *testing.T) {
 				if req.URL.Path == "/v1/version" {
 					return newMockAPIResponse(t, versionData), nil
 				}
-				return nil, http.ErrNotSupported
+				return &http.Response{
+					StatusCode: http.StatusNotFound,
+					Body:       io.NopCloser(strings.NewReader("")),
+					Header:     http.Header{"Content-Type": []string{"application/json"}},
+				}, nil
 			},
 		},
 	}


controllers/llamastackdistribution_controller_test.go
```diff
@@ -312,3 +320,132 @@ func TestReconcile(t *testing.T) {
...
+	// define the data structure for the mock version response
+	versionData := struct {
+		Version string `json:"version"`
+	}{
+		Version: expectedLlamaStackVersionInfo,
+	}
+
+	// create the mock http client that uses our custom roundtripper
+	mockClient := &http.Client{
+		Transport: &mockRoundTripper{
```
rhdedgar, line 378: Could add a timeout here to match the non-test client (if you want to):

```suggestion
	mockClient := &http.Client{
	        Timeout: 5 * time.Second,
		Transport: &mockRoundTripper{
```

rhdedgar (commented):
It looks good! I didn't notice any blockers.

 I had one suggestion just to update the `checkHealth` function to use the injected `r.httpClient` at lines `843` and `852` of `controllers/llamastackdistribution_controller.go` for to match the new convention, but github won't let me add a comment to that range.

alinaryan (approved):
LGTM! Thanks for working on this

Issue #134, opened 2025-08-04
Switch from ConfigMaps to CRs for CA bundles and custom run.yaml data


rhdedgar:
**Background**
Currently, there are multiple ConfigMaps with user-provided data that are watched by the operator for updates. Including CA bundle ConfigMaps and run.yaml data ConfigMaps.

This can potentially create unnecessary resource overhead on very busy Kubernetes clusters that have many ConfigMaps. Especially if they're older clusters that do not have field indexer support.

**Proposed solution**
- [ ] Add an arbitrary data field for run.yaml data (e.g. `spec.server.userConfig.customConfig`), to be used similar to the way ConfigMaps are currently used for this purpose. This should be independent of #117 so it can be used for custom distributions, etc. Where we don't enforce the schema.
- [ ] Add a field for PEM formatted custom CA bundle data (e.g. `spec.server.tlsConfig.caBundle.Data`) 
- [ ] No longer watch CA bundle and run.yaml related ConfigMaps for updates.

Pros

- A LlamaStackDistribution becomes the single source of truth for data that's currently fragmented into 3 pieces (llds, run.yaml, CA bundle).
- Fewer resources to watch for updates, - just the LlamaStackDistribution itself for run.yaml or CA data. (+DSCInitialization midstream later)
- Less time between the time-of-check and the time-of-use for PEM data.

Cons

- Indirection / another layer of abstraction: the relevant data from LlamaStackDistribution will likely need to be turned into an ephemeral ConfigMap in order to be mounted by the pod.
- The midstream implementation may end up with some duplicate ConfigMap data, if copied directly from the DSCInitialization into an operator-managed ephemeral ConfigMap.

rhdedgar:
Closing this Issue now that I was able to obtain a support exception for the older versions of OpenShift. That was the original driving factor behind this Issue.

PR #135, opened 2025-08-04
feat: auto-watch ODH CA bundles for updates

rhdedgar:
Adds a watch for CA bundle ConfigMaps that are referenced by LlamaStackDistributions. 

This will most likely be made obsolete by #134 if we can move away from ConfigMap-provided data, and use the LlamaStackDistribution (+DSCInitialization midstream) for all configuration related to run.yaml and CA bundle data. This should provide a decent holdover until we can reach that state.

### DIFF
--- a/controllers/llamastackdistribution_controller.go
+++ b/controllers/llamastackdistribution_controller.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mfleader:
I think we should add these test cases for your changes

- Unit: auto ODH annotation
  - **Given**: CR without `spec.server.tlsConfig.caBundle`; namespace has `ConfigMap` named `odh-trusted-ca-bundle` with 2 valid PEM keys
  - **When**: `buildPodAnnotations` runs
  - **Then**: pod template has `configmap.hash/odh-ca-bundle` and does not have `configmap.hash/ca-bundle`

- Unit: explicit CA takes precedence
  - **Given**: CR with explicit `spec.server.tlsConfig.caBundle`; namespace may also have `odh-trusted-ca-bundle`
  - **When**: `buildPodAnnotations` runs
  - **Then**: pod template has `configmap.hash/ca-bundle` and does not have `configmap.hash/odh-ca-bundle`

- Unit: deterministic ODH hash
  - **Given**: Two `ConfigMap` objects named `odh-trusted-ca-bundle` with identical key/value pairs but different key insertion orders
  - **When**: `getAutoDetectedCABundleConfigMapHash` runs for each
  - **Then**: hashes are equal; changing any cert content or the set of keys changes the hash

- Envtest: ODH watch maps to CR and triggers rollout
  - **Given**: CR without explicit CA; `ConfigMap` `odh-trusted-ca-bundle` with valid PEM keys in the same namespace
  - **When**: `findLlamaStackDistributionsForConfigMap` is invoked for that `ConfigMap`, and the CR is reconciled
  - **Then**: the CR is returned by the mapper; the resulting `Deployment` template includes `configmap.hash/odh-ca-bundle`; updating the `ConfigMap` data results in a different annotation value on the `Deployment` template

PR #136, opened 2025-08-05
fix(plugins): handle nil intermediate values in JSON path traversal

mfleader:
Simplify field mutator implementation by leveraging jsonpointer v0.21.2 fix for nil intermediate values during field mutation. Replaces complex manual traversal with direct jsonpointer calls and streamlined path creation logic. Improved error handling through jsonpointer's improved error messages.

Previously worked around: panic: reflect: call of reflect.Value.Type on zero Value.


alinaryan (approved):
LGTM

rhdedgar (approved):
Looks good here, thanks!

PR #138, opened 2025-08-08
feat(CI/CD): stub workflow for CPU-based vLLM image generation

courtneypacheco:
Create a stub workflow to test PR #129.


courtneypacheco:
@mfleader  or @rhdedgar  - can you trigger the workflows here? TIA!

PR #154, opened 2025-09-03
(fix): Remove selectableFields to support older k8s version

VaishnaviHire:


rhdedgar:
I have some of the changes from this PR staged in #156 in order to remove the ConfigMap watching dependency entirely. That should lower resource consumption on busy clusters as well, which will be a consideration for newer clusters after the selectableFields are gone.

Issue #117, opened 2025-07-18
Proposal: Add run.yaml directly into LLSD as a sub-schema


rhuss:
For ease of defining the LLS runtime configuration it might make sense to directly embed a subschema that contains the `run.yaml` content.

Benefits of this approach are:

* More accessible for user as they don't have to look and manage a `ConfigMap` (that probably should be a `Secret` anyway).
* Schema support in IDEs when editing the LLSD CR (with autocompletion in the best case). That's not available when editing a CM
* No issues with watching `ConfigMap`s which already lead to performance issues on larger clusters if not indexed CM index is used
* If are not forces to stat 1:1 with run.yaml we could also optimized for RHOAI use cased. E.g. when switching to a remote db for many providers, we could introduce a global `backendDbConnection:` field that we fan-out in the real run.yaml for the individual providers (as long as upstream doesn't support this field). Danger here is of course to diverge too much.

Cons are:

* Coupling the CR configuration to the upstream `run.yaml` schema which might be an issue when the operator supports multi different LLS versions (i.e. the run.yaml depends on the distribution image that we include). This is difficult challenge if run.yaml is changing often across LLS version, not so a huge issue when run.yaml is stable or only changes in backwards-compatible way.

There might be different shades of gray of still allowing a `useConfig:` field to reference a run.yaml that is stored within a CM as alternative. We would need that for the custom distribution case anyway as we can't know what's in those custom distributions.

Another variation is to not fully support *all* of run.yaml but only that are relevant for our operation.

-----

This is an open discussion to find out how and how much configuration we can exposed already in the LLSD CR to make for a smoother and more flexible user experience.

Please add your thoughts and concerns in the comments.

VaishnaviHire:
@rhuss I guess this is still a requirement?

@rhdedgar Okay if I pick this up or are you looking into this?

VaishnaviHire:
@rhuss @rhdedgar Couple of follow-ups here

- Should we keep the option for users to add a configmap optionally? This will allow us to provide users with `managed` vs `unmanaged` configuration.  
- The server config will look as follows -
```
 server:
    distribution:
    
    # OPTION A: Managed configuration with schema validation
    config:
      server:
        port: 8321
      
     # We no longer need users to have knowledge of any env variables.
      providers:
        inference:
        - provider_id: vllm-0
          provider_type: remote::vllm
          config:
            url: "http://vllm-server-service.vllm-dist.svc.cluster.local:11434"
      
[...]

PR #156, opened 2025-09-03
feat: move to llsd for ca bundle and run.yaml config

### LINKED ISSUE #134: Switch from ConfigMaps to CRs for CA bundles and custom run.yaml data

rhdedgar:
With this PR, users no longer need to create separate ConfigMaps for CA bundle and run.yaml configuration data.

- No longer relies on FieldSelectors / Field Indexers for Custom Resources, improving support for older versions of Kubernetes (<1.31) and OpenShift (<4.17)
- Improved performance on clusters that have many ConfigMaps, especially on older clusters.

Closes #134 

Closes #135 

Closes RHAIENG-662


leseb (commented):
I don't understand the rationale for this change and the downside of watching the CM but this feels like a bit of a regression. Now users have to edit the CR with the rather tedious PEM content. Some clusters (like OpenShift) automatically generate the CA bundles in a ConfigMap, which makes it much more convenient to just reference them directly.

rhuss (commented):
@rhdedgar thanks for the PR! I think it goes in the right direction, but the original intention for the configuration override with a custom `run.yaml` was:

Add a subschema of `run.yaml` directly in the resource file. It should be reflected in the CRDs schema. It was not meant to include the run.yaml as a single opaque string, but a more typed approach that helps user to create the fields. This is also a means to protect users for fast schema changes in run.yaml as it currently still happens quite a lot. With keeping a sub-schema when can easily add an additional mapping layer for newer version of the run.yaml schema, so that we keep the operator config stable over a longer time.

For a full overwrite, I would still _keep_ the configMap option, but maybe only locally in the same namespace and picking it up only during startup (no watch needed).

Just moving the full run.yaml makes it even harder to maintain (updating a string inlined yaml is even more error prone with the formatting, especially when it has been changed by some k8s mangling)

Not sure if this makes sense, I would love to hear some more opinions on this (and especially _what_ subschema from run.yaml should we allow).

rhuss:
Yeah, I think for the configuration I think we should allow references only to ConfigMaps in the same namespace (config). This should fix the performance issue without being too restrictive.

For the caBundle, I'm unsure how it works in a multi-tenant setup, where the operator runs completely outside any tenant and different tenants have different CAs they trust. (but then, I'm not deep in how the ca bundle is currently used, so there might be an easy solution, too)

VaishnaviHire:
> For the caBundle, I'm unsure how it works in a multi-tenant setup, where the operator runs completely outside any tenant and different tenants have different CAs they trust. (but then, I'm not deep in how the ca bundle is currently used, so there might be an easy solution, too)

We currently reference the ca-bundle configmap in our CR. For multi-tenancy, I think we can just watch the tenant namespace/s? Require users to have the ca-bundle configmap within those namespaces?

rhuss:
@rhdedgar I fully agree that we should keep the same semantics about restarting, regardless of how the configuration is stored. 

For the external, fully-overridable `run.yaml` case, wdyt think about the idea, that the ConfigMaps themselves should be immutable (you can set CMs with a flag immutable), and then require the user to make an update of the reference field in the LLSD pointing to the ConfigMap ? (probably somehting like `customConfig: run-yaml-config-asdff1234`, changing the suffix for every update). Something what kustomize does, too. 

The benefit of this approach is that the reconciled _Deployment_ resource differs from the original one (change in the configmap name), so going to restart the deployment. If you only change the configMap but not the DeploymentSpec, usually nothing happens. Of course, any ConfigMap mounted into a Pod reflects the new content, but the process within the Pod's container needs to detect that (via a file watch) and need to do a hot-reload (without restart). That would be the best solution anyways, but LLS is not capable of doing this.

Using immutable configMaps also the benefit of having any way for history and allows rollback, too.

tl;dr - I probably would favor such a solution:

* Keep the ConfigMap reference and don't include a full `run.yaml` (that really is a bad and fragile user experience, mainly when run.yaml can get large. Additionally, you need to perform special escaping, such as $ -> $$, due to the special YAML semantics. It's really not trivial.
* Encourage people who use an external configMap to use either kustomize to manage their LLSDs or recommend immutable configmaps to trigger restarts.
* Focus on #117 _in addition_ to allow a smoother and more stable user experience for simpler use cases, like adding additional models or tools.

rhdedgar:
Closing this PR, as I was able to obtain a support exception for the older versions of OpenShift.

PR #160, opened 2025-09-10
test: remove ollama distribution testing, replace with starter

derekhiggins:
test: remove ollama distribution testing, replace with starter
    
    This focuses testing on the actively maintained starter distribution
    while removing support for the older ollama image that's no longer updated.
    
    Also intrduces a list of distro images (currently only 1) that can be tested.
    
---
    test: combine creation and deletion into single lifecycle test
    
    Restructure the creation test to run creation followed by deletion in a single
    test suite. This enables testing multiple distribution images sequentially
    with proper cleanup between runs.
    
    Changes:
    - Add TestCreationDeletionSuite that runs creation then deletion
    - Replace TestDeletionSuite with runDeletionTests helper function
    - Simplify test structure while maintaining all existing functionality

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -3,6 +3,9 @@ package e2e
 
 import (
 	"testing"
+
+	"github.com/llamastack/llama-stack-k8s-operator/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestE2E(t *testing.T) {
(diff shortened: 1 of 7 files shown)


tests/e2e/test_utils.go
```diff
@@ -28,6 +28,8 @@ import (
 )
 
 const (
+	starterDistType      = "starter"
+	ollamaDistType       = "ollama"
```
leseb, line 32: ollama does not really exist anymore as a distro, I'm not sure it's worth keeping it... Can we remove it in favor of starter?

derekhiggins: Can do, I was under the impression that the llsd operator should maintain backwards compat with older distro images but I'm happy to remove it if we don't need it.

leseb: ollama distro hasn't been updated since 0.2.9 😓 so yeah let's remove it, thanks!

rhdedgar (approved):
Looks good to me! I tested this with `config/samples/example-with-configmap.yaml` and `config/samples/example-with-ca-bundle.yaml`, both run as expected.

tests/e2e/e2e_test.go
```diff
@@ -3,31 +3,58 @@ package e2e
...
-	creationFailed := false
-
-	// Run creation tests
-	t.Run("creation", func(t *testing.T) {
-		TestCreationSuite(t)
-		creationFailed = t.Failed()
-	})
+	// Run combined creation and deletion tests for multiple distributions
+	// starter: newer image currently being actively updated
+	distributions := []string{"starter"}
```
leseb, line 17: Would be nice to be able to detect the distro instead of having to change that code. Improvement for later I suppose.

leseb (approved):

PR #161, opened 2025-09-13
fix: init pod run perms

rhdedgar:
Setting FSGroup to convey the need for a more permissive scc.

Closes RHAIENG-999.


rhdedgar:
It's possible the `RunAsUser` setting could even be removed in the future, as the `FSGroup: 0` setting should provide the necessary volume access.

derekhiggins:
So we're dropping some privileges from the init container while mounting storage in as the 'root' group so that containers can always write to volumes?

Would you mind putting a little context on why this is needed in here (as not everybody can see what RHAIENG-999 is )

PR #163, opened 2025-09-17
chore: remove review-requested=0 condition to prevent auto-merge blocking

derekhiggins:
Allows Mergify to auto-merge PRs even when there are pending review requests, as long as the minimum approval threshold is met and no changes are requested.

This prevents situations where PRs with sufficient approvals get blocked because some requested reviewers haven't responded yet.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -4,7 +4,6 @@ pull_request_rules:
     Automatic merge of PRs to main
   conditions:
     - "#approved-reviews-by>=1"
-    - "#review-requested=0"
     - "#changes-requested-reviews-by=0"
     - base=main
     - label!=do-not-merge


nathan-weinberg:
I'm in favor of this, if we can do a 2-approval minimum.

nathan-weinberg:
> > Let's please use GitHub Teams or some other mechanism then so if we want groups of people looking at things (Dev, QE, Docs, etc) we can add them more easily
> 
> hmm, I'm not sure if we can do this, they would need to be setup at the org level (I think), or is there somewhere they can be setup in the repo?

Alas I think you are right - I'd be happy to do this if I can get the permissions, otherwise we may need to indeed cut down on the codeowners and do manual additions of individuals

VaishnaviHire:
This is stale, current config has been already updated

PR #165, opened 2025-09-22
(chore): Update docker action tag

VaishnaviHire:
This commit updates the docker action tag to unblock the release workflow

Existing error - https://github.com/llamastack/llama-stack-k8s-operator/actions/runs/17920005053/job/50952690570


derekhiggins (approved):
lgtm, looks like the old hash was from `actions/setup-python`
new has is found here 
https://github.com/docker/login-action/commits/releases/v2

PR #169, opened 2025-10-01
feat(controller): ConfigMap image mapping override

derekhiggins:
feat: ConfigMap image mapping overrides for LLS Distro
    
    Implements a mechanism for the Llama Stack Operator to read and apply
    LLS Distribution image updates from a ConfigMap, enabling independent patching
    for security fixes or bug fixes without requiring a new LLS Operator.
    
    - Add ImageMappingOverrides field to LlamaStackDistributionReconciler
    - Implement parseImageMappingOverrides() to read image-overrides from ConfigMap
    - Support symbolic name mapping (e.g., rh-dev) to specific images
    - Included unit tests
    
    The operator now reads image overrides from the 'image-overrides' key in the
    operator ConfigMap, supporting YAML format with version-to-image mappings.
    Overrides take precedence over default distribution images and are refreshed
    on each reconciler initialization.
    
    Closes: RHAIENG-1079
    Signed-off-by: Derek Higgins <derekh@redhat.com>

-----

    Fix operator config ConfigMap watching
    
    o Add operator config detection in configMapUpdatePredicate
    o Update feature flags when operator config changes
    o Trigger reconciliation of all LlamaStackDistributions on operator config change

-----
Also

     fix: Replace deprecated corev1.Endpoints with discoveryv1.EndpointSlice 

Fix for deprecated Endpoints Generated by cursor but output tested here [here ](https://productionresultssa7.blob.core.windows.net/actions-results/d704a1cd-fe19-400e-8d64-67fbef9e8d44/workflow-job-run-7a18097a-388d-593a-be57-392c834218f2/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-10-01T11%3A47%3A59Z&sig=0Bu8kxdp29iP%2FexZTOUpyo2i0hk5woMASdnCv2juakM%3D&ske=2025-10-01T22%3A02%3A40Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-10-01T10%3A02%3A40Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-10-01T11%3A37%3A54Z&sv=2025-11-05) , search for the string "Found 1 pods in namespace"


config/crd/bases/llamastack.io_llamastackdistributions.yaml
```diff
@@ -107,8 +107,12 @@ spec:
                                       description: The key to select.
                                       type: string
                                     name:
+                                      default: ""
                                       description: |-
                                         Name of the referent.
+                                        This field is effectively required, but due to backwards compatibility is
+                                        allowed to be empty. Instances of this type with an empty value here are
+                                        almost certainly wrong.
```
rhuss, line 115: You mean, the name of the referenced ConfigMap ? Maybe we can add this here instead of just `Name of referent.` (and if so, shouldn't it be `reference` as the referent is the operator itself ?)

config/crd/bases/llamastack.io_llamastackdistributions.yaml
```diff
@@ -169,8 +173,12 @@ spec:
                                         from.  Must be a valid secret key.
                                       type: string
                                     name:
+                                      default: ""
                                       description: |-
                                         Name of the referent.
+                                        This field is effectively required, but due to backwards compatibility is
```
rhuss, line 179: Same as above. Just out of curiosity, why was this name ever allowed to be empty ?

config/crd/bases/llamastack.io_llamastackdistributions.yaml
```diff
@@ -214,6 +222,12 @@ spec:
                                     the Pod where this field is used. It makes that resource available
                                     inside a container.
                                   type: string
+                                request:
+                                  description: |-
+                                    Request is the name chosen for a request in the referenced claim.
+                                    If empty, everything from the claim is made available, otherwise
+                                    only the result of this request.
+                                  type: string
```
rhuss, line 230: This description is not clear to me. What is "the result of this request" ?

config/crd/bases/llamastack.io_llamastackdistributions.yaml
rhuss: It looks like that the descriptions here are picked up from somewhere else, also not necessarily coming from the operator code. How have those been added ?

config/rbac/role.yaml
```diff
@@ -87,6 +87,14 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - clusterserviceversions
```
rhuss, line 93: this is a tricky permission as this is a cluster-wide permissions and a platform engineer or "LLS admin" might not have this right and so this might be an unwanted elevation of permissions.

From the description of the PR I understand that this is needed to get the version of RHODS. I'm not sure, why we do need this and it's not really appropriate for an upstream operator to have _any_ OpenShift references (including references to RHODS).

The operator should be also independent of OpenDataHub, and be able to be installed standalone on Kubernetes.

I wouldn't add this dependency here to this OpenShift specific resource (well, its operator specific but let's face it, this is more or less only used by OpenShift in the wild)

controllers/llamastackdistribution_controller.go
```diff
@@ -80,11 +81,19 @@ const (
...
+// Operator ConfigMap Watching Feature:
+// This reconciler also watches for changes to the operator configuration ConfigMap. When the operator
+// config changes, it triggers reconciliation of all LlamaStackDistribution resources.
 type LlamaStackDistributionReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 	// Feature flags
 	EnableNetworkPolicy bool
+	// RHODS version
+	RHODSVersion string
```
rhuss, line 94: As said, we can't used a reference to RHODS in the upstream operator. We could patch it in midstream, but maybe we can abstract this away to make it more general ? I'm not yet clear why we need this version.

controllers/llamastackdistribution_controller.go
```diff
@@ -458,6 +467,40 @@ func (r *LlamaStackDistributionReconciler) configMapUpdatePredicate(e event.Upda
...
+	if err != nil {
+		return false
+	}
+
+	if configMap.Name != operatorConfigData || configMap.Namespace != operatorNamespace {
+		return false
+	}
+
+	// Update feature flags
+	EnableNetworkPolicy, err := parseFeatureFlags(configMap.Data)
```
rhuss, line 491: Why is network policy enabled/disable via a feature flag (which is used for enabling/disabling experimental features which become eventually permanent features (or dropped)).

For me it looks like the usage of NetworkPolicies is an admin decision so that it shoudl be part of the regular configuration. A feature flag could handle if this configuration has any effect, but the overall switch network policy yes/no should be part of LLSD.

rhuss (changes requested):
Thanks for the PR, great stuff :). Here's my first round of review.

My main points are:

* We must not reference RHODS in this upstream operator. Why and where do you need the RHODS version ? (i can't really see why this is needed). This is really important.
* Probably not related to this PR, nevertheless: The enabling of network policies should not be a 'feature flag' (which is about enabling/disabling feature that has its own configuration), but a 'configuration option' (you can combine both, feature flag and configuration option but you should not 'misuse' a feature flag as configuration of a permanent feature)

README.md
```diff
@@ -104,6 +104,48 @@ Example to create a run.yaml ConfigMap, and a LlamaStackDistribution that refere
...
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: llama-stack-operator-config
+  namespace: llama-stack-k8s-operator-system
+data:
+  image-overrides: |
+    rh-dev: quay.io/rhoai/rhoai-fbc-fragment:rhoai-2.25@sha256:3bc98555
```
VaishnaviHire: I think we should just use the available upstream distributions for example docs

tests/e2e/test_utils.go
```diff
@@ -313,34 +314,77 @@ func logPodDetails(t *testing.T, testenv *TestEnvironment, namespace string) {
...
 	}
 
-	t.Logf("🔗 Service %s endpoints:", serviceName)
-	for i, subset := range endpoints.Subsets {
-		t.Logf("  Subset %d:", i)
-		// Ready addresses indicate pods that passed health checks and can receive traffic
-		t.Logf("    Ready addresses: %d", len(subset.Addresses))
-		for _, addr := range subset.Addresses {
-			t.Logf("      - %s", addr.IP)
+	t.Logf("🔗 Service %s endpoint slices:", serviceName)
```
VaishnaviHire: Are these changes related to configmap?

README.md
```diff
@@ -104,6 +104,48 @@ Example to create a run.yaml ConfigMap, and a LlamaStackDistribution that refere
...
+    starter: quay.io/custom/llama-stack:starter
+```
+
+### Configuration Format
+
+Use the distribution name directly as the key (e.g., `rh-dev`, `starter`). The operator will apply these overrides automatically
+
+### How It Works
```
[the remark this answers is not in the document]
derekhiggins: I'm not sure what you mean here, do you want me to add this to the usage section below ?

README.md
```diff
@@ -104,6 +104,48 @@ Example to create a run.yaml ConfigMap, and a LlamaStackDistribution that refere
...
+
+1. The operator reads image overrides from the `image-overrides` key in the operator ConfigMap
+2. Overrides are parsed as YAML with distribution-name-to-image mappings
+3. When deploying a LlamaStackDistribution, the operator checks for overrides matching the distribution name
+4. If an override exists, it uses the specified image instead of the default distribution image
+5. Changes to the ConfigMap automatically trigger reconciliation of all LlamaStackDistribution resources
+
+### Example Usage
+
+To update the LLS Distribution image for all `rh-dev` distributions:
```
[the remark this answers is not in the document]
derekhiggins: done

controllers/llamastackdistribution_controller.go
```diff
@@ -1632,53 +1693,92 @@ func NewLlamaStackDistributionReconciler(ctx context.Context, client client.Clie
...
+func ParseImageMappingOverrides(configMapData map[string]string) map[string]string {
+	imageMappingOverrides := make(map[string]string)
+
+	// Look for the image-overrides key in the ConfigMap data
+	if overridesYAML, exists := configMapData["image-overrides"]; exists {
+		// Parse the YAML content
+		var overrides map[string]string
+		if err := yaml.Unmarshal([]byte(overridesYAML), &overrides); err != nil {
+			// Log error but continue with empty overrides
+			fmt.Printf("failed to parse image-overrides YAML: %v\n", err)
```
rhdedgar: Could this be updated to use `logger.Error` instead, if it's a breaking event? Day 2 Operations partially focuses on structured logging, like using `logger.V(1).Info` for debug info, or `logger.Info` for general information.

controllers/llamastackdistribution_controller.go
```diff
@@ -1632,53 +1693,92 @@ func NewLlamaStackDistributionReconciler(ctx context.Context, client client.Clie
...
+			return imageMappingOverrides
 		}
 
-		if err = client.Create(ctx, configMap); err != nil {
-			return nil, fmt.Errorf("failed to create ConfigMap: %w", err)
+		// Copy the parsed overrides to our result map
+		for version, image := range overrides {
+			imageMappingOverrides[version] = image
 		}
```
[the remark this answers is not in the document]
derekhiggins: I've added 'ParseReference' to validate the image can be parsed but stopped short of only allowing specific repositories as users may have their own that they want to use.

VaishnaviHire:
@rhdedgar @rhuss Is this PR okay to merge?

rhdedgar (approved):

derekhiggins:
rebased to prompt CI into running

VaishnaviHire:
@mergify rebase

PR #170, opened 2025-10-03
feat: add fips-compliant annotation to CSV

rhdedgar:
Adding `features.operators.openshift.io/fips-compliant: "true"` annotation to the CSV as per the FIPS Operator checklist.

Closes RHAIENG-1325.


VaishnaviHire:
@mergify rebase

PR #171, opened 2025-10-06
chore: use the new CLI to run the server

leseb:
0.3.0 with https://github.com/llamastack/llama-stack/pull/3625 forces us to use "llama stack run" and the server module doesn't execute the server anymore.


VaishnaviHire:
/lgtm

+1 for moving it to distribution image.

PR #172, opened 2025-10-07
fix: allow non-admins to use storage

rhdedgar:
Addresses behavior common to issues like RHAIENG-1139 and RHAIENG-999, where:

1. Users creating LlamaStackDistributions with the same name in multiple namespaces get an scc permissions error.
2. Unprivileged users trying to create LlamaStackDistributions with the `storage` option set were getting the same permissions error.

Opened for discussion, on if this is the approach we'd like to take to address the scope issue.

### DIFF
--- a/pkg/deploy/kustomizer.go
+++ b/pkg/deploy/kustomizer.go
@@ -28,6 +28,24 @@ import (
 	yamlpkg "sigs.k8s.io/yaml"
 )
 
+const (
+	// Resource naming constants.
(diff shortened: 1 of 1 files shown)


VaishnaviHire:
@rhdedgar Can you confirm if these changes are still needed after https://github.com/llamastack/llama-stack-k8s-operator/pull/173 ?

rhdedgar:
Closing now that #173 has been merged.

PR #177, opened 2025-11-07
Migrate away from initContainers for CA Bundle operations

rhdedgar:
To provide better support for disconnected environments, the operator's initContainers are being removed.

This PR removes the initContainer that performs CA Bundle concatenation, and now processes cert data from ConfigMaps directly before mounting the final result into the llama-stack container.

Validation steps are performed to ensure that only legitimate certificate data is being passed to the llama-stack process in the container. Other basic size /number validation is done to ensure the operator stays within its resource limits. 

Startup script additions were avoided this time, so that intentions like RHAIENG-1438 can be carried out without additional logic overhead.

Supersedes #174
Closes RHAIENG-1383

### DIFF
--- a/tests/e2e/tls_test.go
+++ b/tests/e2e/tls_test.go
@@ -100,6 +100,10 @@ func testLlamaStackWithCABundle(t *testing.T) {
 	err = waitForDeploymentCreation(t, llsTestNS, "llamastack-with-config", 3*time.Minute)
 	require.NoError(t, err, "LlamaStack deployment should be created by operator")
 
+	// Wait for pods to be running and ready
+	err = WaitForPodsReady(t, TestEnv, llsTestNS, "llamastack-with-config", 5*time.Minute)
+	require.NoError(t, err, "LlamaStack pods should be running and ready")
+
 	// Verify certificate volumes are mounted correctly
 	err = verifyCertificateMounts(t, llsTestNS, "llamastack-with-config")
 	require.NoError(t, err, "Certificate volumes should be mounted correctly")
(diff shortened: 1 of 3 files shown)


controllers/llamastackdistribution_controller.go
```diff
@@ -1190,34 +1227,315 @@ func (r *LlamaStackDistributionReconciler) getConfigMapHash(ctx context.Context,
...
+		if err := collector.add(certs, size, count, configMap.Name, key); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *LlamaStackDistributionReconciler) processODHConfigMapKeys(configMap *corev1.ConfigMap, keys []string, collector *certificateCollector) error {
+	for _, key := range keys {
```
VaishnaviHire: Is there a reason to explicitly have odh specific function here instead of just using the above function?

rhdedgar: This function was intended to be different from `processConfigMapKeys` in a couple ways.

For `processConfigMapKeys`, if the user specifies a ConfigMap that doesn't exist or a Key that has formatting issues, the deployment should fail until those issues are resolved by the user.

If the ODH bundle has some kind of issue, it may not be something that can be resolved by the user if they don't also manage the ODH instance. So the `processODHConfigMapKeys` function only warns them about it and then continues.

tests/e2e/tls_test.go
```diff
@@ -419,7 +424,7 @@ func verifyEnvironmentVariables(t *testing.T, namespace, name string) error {
 	// Check for TLS-related environment variables
 	tlsEnvVarsFound := 0
 	expectedEnvVars := map[string]string{
-		"VLLM_TLS_VERIFY": controllers.CABundleMountPath,
+		"SSL_CERT_FILE": controllers.ManagedCABundleFilePath,
 	}
```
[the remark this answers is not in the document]
rhdedgar: Sure, I've updated this PR with a new commit to check for pods in the `Ready` status.

PR #221, opened 2026-01-06
Add support for Network config

### LINKED ISSUE #212: Introduce network config in LlamaStackDistribution
- As a platform admin, I want to allow specific namespaces (explicitly or by label) to reach LlamaStack services, and optionally expose ingress/route for external access.
- Add a network config to LlamaStackDistribution to control ingress exposure and namespace-based access for NetworkPolicies (NP).
- Default to lockdown (only LLSD namespace + operator pod) but allow widening access declaratively.
- Explicit namespace and label-based controls reduce manual NP edits. 

**Implementation Details:**

- `spec.network.exposeRoute: bool` - default false; when true, create ingress/route for external access.
- `spec.network.allowedFrom.namespaces: []string` -  explicit list of namespaces allowed; "*" means all namespaces.
- `spec.network.allowedFrom.labels: []string` - list of namespace label selectors (match any) that are allowed; requires watching namespaces and updating NPs dynamically.
- Egress: remain open (unchanged).
```
network:
  exposeRoute: false 
  allowedFrom:
    namespaces:   
    - nsA
    - nsB
    labels:          
    - myproject/lls-allowed
```

**Acceptance criteria**

- CRD schema updated with network section (validation + defaults).
- Controller reconciles NetworkPolicies to honor namespace list and label selectors; continues to allow LLSD namespace + operator.
- Optional ingress/route is created/updated when exposeRoute toggled.
- Docs + sample CR updated to show the new fields and defaults.

VaishnaviHire:
Fixes #212 

I have kept the global `EnableNetworkPolicy` as is, to allow users to add/remove network policies for all lls instances at once.


controllers/network_resources.go
```diff
@@ -0,0 +1,219 @@
...
+		return fmt.Errorf("failed to get Ingress: %w", getErr)
+	}
+
+	if !metav1.IsControlledBy(existing, instance) {
+		logger.V(1).Info("Ingress not owned by this instance, skipping update", "name", ingressName)
+		return nil
+	}
+
+	ingress.ResourceVersion = existing.ResourceVersion
+	if updateErr := r.Update(ctx, ingress); updateErr != nil {
```
leseb: ditto

PR #223, opened 2026-01-08
Update run.yaml references to config.yaml

VaishnaviHire:
Remove run.yaml references and replace with config.yaml

### DIFF
--- a/config/samples/example-with-ca-bundle.yaml
+++ b/config/samples/example-with-ca-bundle.yaml
@@ -11,13 +11,13 @@ data:
     - inference
     providers:
       inference:
-      - provider_id: vllm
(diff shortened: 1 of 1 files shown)


rhdedgar (changes requested):
I would say that if these remaining references get updated, we should be good to go:

```
config/samples/example-with-ca-bundle.yaml:6:  run.yaml: |
config/samples/example-with-configmap.yaml:6:  run.yaml: |
controllers/llamastackdistribution_controller_test.go:163:			"run.yaml": `version: '2'
controllers/llamastackdistribution_controller_test.go:209:	configMap.Data["run.yaml"] = `version: '2'
```

PR #227, opened 2026-01-21
fix: Update networkpolicy to remove podSelectors

VaishnaviHire:
- With NetworkPolicy enabled, external traffic through OpenShift Routes is blocked even when the correct namespaces/labels are configured in allowedFrom. This is because the NetworkPolicy includes an explicit `podSelector: {}` in the ingress rules, which doesn't match HostNetwork traffic from the OpenShift router.
- Remove references to managed OpenShift Route. LLS operator only creates Ingress resource.

### DIFF
--- a/config/crd/bases/llamastack.io_llamastackdistributions.yaml
+++ b/config/crd/bases/llamastack.io_llamastackdistributions.yaml
@@ -84,7 +84,7 @@ spec:
                   exposeRoute:
                     default: false
                     description: |-
-                      ExposeRoute when true, creates an Ingress (or OpenShift Route) for external access.
+                      ExposeRoute when true, creates an Ingress for external access.
                       Default is false (internal access only).
                     type: boolean
                 type: object
--- a/api/v1alpha1/llamastackdistribution_types.go
+++ b/api/v1alpha1/llamastackdistribution_types.go
@@ -82,7 +82,7 @@ type LlamaStackDistributionSpec struct {
 
 // NetworkSpec defines network access controls for the LlamaStack service.
 type NetworkSpec struct {
-	// ExposeRoute when true, creates an Ingress (or OpenShift Route) for external access.
+	// ExposeRoute when true, creates an Ingress for external access.
 	// Default is false (internal access only).
 	// +optional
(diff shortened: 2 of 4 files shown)


config/crd/bases/llamastack.io_llamastackdistributions.yaml
```diff
@@ -84,7 +84,7 @@ spec:
                   exposeRoute:
                     default: false
                     description: |-
-                      ExposeRoute when true, creates an Ingress (or OpenShift Route) for external access.
```
jgarciao: Instead of using exposeRoute as boolean, could you consider using the "Type + Config" Pattern ?

Examples (created with help from a chatbot):
```
spec:
  network:
    exposure:
      type: Route
      host: llama.apps.my-cluster.com
      route:
        termination: edge
```

```
spec:
  network:
    exposure:
      type: Ingress
      host: llama.example.com
      ingress:
        className: nginx
        annotations:
          cert-manager.io/cluster-issuer: "letsencrypt-prod"
```

```
  network:
    type: object
    description: "Network configuration for the Llama Stack."
    properties:
      exposure:
        type: object
        description: "Defines how the Llama Stack service is exposed externally."
        properties:
          type:
            type: string
            default: None
            enum:
              - None
              - Ingress
              - Route
            description: |
              Type of external exposure to create.
              None = internal only
              Ingress = create a standard Kubernetes Ingress
              Route = create an OpenShift Route
          host:
            type: string
            description: "The external hostname for the service (used by both Ingress and Route)."
          ingress:
            type: object
            description: "Optional overrides specifically for Kubernetes Ingress."
            properties:
              className:
                type: string
                description: "The ingressClassName to use."
              annotations:
                type: object
                additionalProperties:
                  type: string
                description: "Additional annotations for the Ingress resource."
          route:
            type: object
            description: "Optional overrides specifically for OpenShift Routes."
            properties:
              termination:
                type: string
                enum: [edge, passthrough, reencrypt]
                description: "TLS termination type for the OpenShift Route."
```

VaishnaviHire:
@mergify rebase

PR #246, opened 2026-02-11
feat: add default env var for kube auth

rhdedgar:
Adding `OPENSHIFT_SERVER_API_URL` as a default LlamaStackDistribution env var, so it doesn't need to be specified manually in the CR. This allows llama-stack servers to validate user tokens against the OpenShift API server for Kubernetes-based RBAC authentication.

The change supports the crimson dashboard feature request at https://issues.redhat.com/browse/RHOAIENG-34370.

### DIFF
--- a/controllers/resource_helper.go
+++ b/controllers/resource_helper.go
@@ -300,7 +300,16 @@ func configureContainerEnvironment(ctx context.Context, r *LlamaStackDistributio
 		},
 	)
 
-	// Finally, add the user provided env vars
+	// Add OpenShift/Kubernetes API server URL for RBAC authentication.
+	// This enables the Kubernetes auth provider in llama-stack to validate tokens
+	// against the API server for access control.
+	// Default to in-cluster service URL; can be overridden via user-provided env vars.
+	container.Env = append(container.Env, corev1.EnvVar{
+		Name:  "OPENSHIFT_SERVER_API_URL",
+		Value: "https://kubernetes.default.svc",
+	})
+
+	// Finally, add the user provided env vars (these take precedence over defaults)
 	container.Env = append(container.Env, instance.Spec.Server.ContainerSpec.Env...)
 }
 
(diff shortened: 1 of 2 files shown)


controllers/resource_helper.go
```diff
@@ -300,7 +300,16 @@ func configureContainerEnvironment(ctx context.Context, r *LlamaStackDistributio
...
 	)
 
-	// Finally, add the user provided env vars
+	// Add OpenShift/Kubernetes API server URL for RBAC authentication.
+	// This enables the Kubernetes auth provider in llama-stack to validate tokens
+	// against the API server for access control.
+	// Default to in-cluster service URL; can be overridden via user-provided env vars.
+	container.Env = append(container.Env, corev1.EnvVar{
+		Name:  "OPENSHIFT_SERVER_API_URL",
+		Value: "https://kubernetes.default.svc",
```
derekhiggins: I'm wondering why this is needed? according to the lls code this what the provider uses as a default 
https://github.com/llamastack/llama-stack/blob/5f66b52061cce554efefd12eb6536965aba33bc8/src/llama_stack/core/datatypes.py#L291

rhdedgar: Hi, this env var won't be used by llama-stack code directly, but rather the odh-dashboard/BFF, which generates the llama-stack config.yaml file at runtime. 

When RBAC is enabled there, the BFF writes `api_server_url: "${env.OPENSHIFT_SERVER_API_URL:=https://kubernetes.default.svc}"` into the generated config. Llama-stack's `replace_env_vars()` then resolves that reference from the container's environment.

I think the llama-stack default referenced by [datatypes.py#L291](https://github.com/llamastack/llama-stack/blob/5f66b52061cce554efefd12eb6536965aba33bc8/src/llama_stack/core/datatypes.py#L291) only applies when `api_server_url` is omitted entirely from the config YAML. But since the BFF explicitly sets the field using an env var reference, that default is never reached, and the value has to come from the env var or its inline fallback.

Maybe I should take it one step further, and have the operator detect the OpenShift API directly server URL (e.g., from `infrastructure.config.openshift.io`) and set it dynamically. Or revisit after the BFF PR merges.

derekhiggins: Sorry, I still don't get it, if BFF writes `api_server_url: "${env.OPENSHIFT_SERVER_API_URL:=https://kubernetes.default.svc}` in config.yaml then you are correct that the default in datatypes.py#L291 does not apply. But in this case BFF is supplying a inline fallback which is the same as OPENSHIFT_SERVER_API_URL so OPENSHIFT_SERVER_API_URL is redundant.

That said if I'm misunderstanding something and this is required, could it be renamed to KUBERNETES_API_SERVER_URL(or somthing similar) or maybe it should be midstream only?

PR #280, opened 2026-04-02
docs: document ENABLE_INLINE_MILVUS limitations in README

mfleader:
[RHAIENG-3916](https://issues.redhat.com/browse/RHAIENG-3916)

### Description

Companion to opendatahub-io/llama-stack-distribution#318.

Documents that `inline::milvus` (enabled via `ENABLE_INLINE_MILVUS`)
is only supported in single-worker, single-replica deployments. The
operator does not manage or validate this env var.

### How Has This Been Tested?

```bash
make test
```

### Test Impact

None. Documentation only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ spec:
 ```
 3. Verify the server pod is running in the user defined namespace.
 
+### Local Vector Storage (inline::milvus)
+
+To enable the `inline::milvus` local vector storage provider, set `ENABLE_INLINE_MILVUS` in `spec.server.containerSpec.env`. This is only supported in single-worker, single-replica deployments. Milvus-Lite uses SQLite internally and does not support concurrent access from multiple processes.
+
 ### Using a ConfigMap for config.yaml configuration
 
 A ConfigMap can be used to store config.yaml configuration for each LlamaStackDistribution.


controllers/resource_helper.go
```diff
@@ -284,6 +284,18 @@ func configureContainerEnvironment(ctx context.Context, r *LlamaStackDistributio
 		})
 	}
 
+	// Enable inline::milvus only when there is no multi-process contention on the
+	// local milvus.db file. Milvus-Lite uses SQLite internally, which only supports
+	// single-process access. Multiple processes opening the same file causes a crash.
+	// This happens with workers > 1 (forked processes in one pod) or replicas > 1
+	// with persistent storage (multiple pods mounting the same PVC).
+	if workers <= 1 && (instance.Spec.Replicas <= 1 || instance.Spec.Server.Storage == nil) {
```
derekhiggins: I'd nearly do the opposite here, i.e. don't touch ENABLE_INLINE_MILVUS at all, but if the user has enabled it (by specifying ENABLE_INLINE_MILVUS in their llsd crd) then output some kind of error/warning that the setup isn't supported. 

If we document that we only support it in certain cases, then let the user decided if they want it or not.

but lets see what others think.

VaishnaviHire: I agree with this. Operator should not manage/set env config for lls distribution

controllers/resource_helper_test.go
```diff
@@ -70,6 +70,7 @@ func TestBuildContainerSpec(t *testing.T) {
 				}},
 				Env: []corev1.EnvVar{
 					{Name: "HF_HOME", Value: "/.llama"},
+
```
VaishnaviHire: nit: can remove this new line

controllers/resource_helper_test.go
```diff
@@ -118,6 +119,7 @@ func TestBuildContainerSpec(t *testing.T) {
 				},
 				Env: []corev1.EnvVar{
 					{Name: "HF_HOME", Value: "/custom/path"},
+
```
VaishnaviHire: same as above

controllers/resource_helper.go
```diff
@@ -763,3 +763,31 @@ func copyIntOrString(value *intstr.IntOrString) *intstr.IntOrString {
...
+	}
+
+	workers, _ := getEffectiveWorkers(instance)
+	multiWorker := workers > 1
+	multiReplicaWithStorage := instance.Spec.Replicas > 1 && instance.Spec.Server.Storage != nil
+
+	if multiWorker || multiReplicaWithStorage {
+		return "ENABLE_INLINE_MILVUS is set but the deployment topology does not support it. " +
+			"Milvus-Lite uses SQLite which only supports single-process access. " +
+			"Workers > 1 or replicas > 1 with persistent storage will cause crashes."
```
derekhiggins: AIUI even without persistent storage this wont work
data would be added to the vector store on one replica and then not available to any requests that hit alternative replica's. So inline vector store doesn't work with replica > 1 

It would be the same situation with inine sqlite-vec , faiss, localfs (maybe others..) , unless we plan on flagging warnings for all config scenarios that have problems we might be better off just putting this in the docs.

VaishnaviHire (approved):

derekhiggins (approved):
lgtm, thanks

PR #282, opened 2026-04-10
fix: Reset feature flags to default on upgrade

VaishnaviHire:
- Introduce ensureDefaultFeatureFlags, which parses the existing feature flags YAML and fills in only missing keys with their compiled defaults, preserving any user-customized values.

- Add a ResetOnUpgrade constant in pkg/featureflags (default false) that gates the full-reset behavior behind a build-time flag. When set to true (midstream), all flags are reset to defaults on operator version change via the existing needsConfigMapUpdate version-detection logic

### DIFF
--- a/controllers/llamastackdistribution_controller.go
+++ b/controllers/llamastackdistribution_controller.go
@@ -1541,17 +1541,7 @@ func initializeOperatorConfigMap(ctx context.Context, c client.Client, operatorN
 
 	err := c.Get(ctx, configMapName, configMap)
 	if err == nil {
-		// Ensure the watch label exists (upgrade path)
-		if configMap.Labels == nil || configMap.Labels[WatchLabelKey] != WatchLabelValue {
-			if configMap.Labels == nil {
-				configMap.Labels = make(map[string]string)
-			}
-			configMap.Labels[WatchLabelKey] = WatchLabelValue
-			if updateErr := c.Update(ctx, configMap); updateErr != nil {
-				return nil, fmt.Errorf("failed to add watch label to operator config ConfigMap: %w", updateErr)
-			}
-		}
-		return configMap, nil
+		return upgradeOperatorConfigMap(ctx, c, configMap, configMapName)
 	}
 
 	if !k8serrors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


controllers/llamastackdistribution_controller.go
```diff
@@ -1571,6 +1561,68 @@ func initializeOperatorConfigMap(ctx context.Context, c client.Client, operatorN
...
+	if needsConfigMapUpdate(ctx, c) {
+		newConfigMap, genErr := createDefaultConfigMap(configMapName)
+		if genErr != nil {
+			return nil, fmt.Errorf("failed to generate default configMap: %w", genErr)
+		}
+		if configMap.Data == nil {
+			configMap.Data = make(map[string]string)
+		}
+		// Only update the featureFlags key, preserve other keys (like image-overrides)
+		configMap.Data[featureflags.FeatureFlagsKey] = newConfigMap.Data[featureflags.FeatureFlagsKey]
```
derekhiggins, line 1591: This will unconditionally reset all feature flags to defaults and potentially break anyone who has set them for a specific reason. Shouldn't we just set them to defaults if they don't exist?

If we only set missing flags, is the version check needed at all? We could just always ensure each featureFlags key exists.

derekhiggins (commented):
See comment inline, also while looking into this I see the operator version in RHOAI doesn't seem correct, any idea why this is?

```
dhiggins@dhiggins-thinkpadp1gen7:~/workarea/ls_claude/lls-showroom/llama-stack-showroom$ oc get csv -n redhat-ods-operator | grep rhods
rhods-operator.3.4.0          Red Hat OpenShift AI               3.4.0                                   Succeeded
dhiggins@dhiggins-thinkpadp1gen7:~/workarea/ls_claude/lls-showroom/llama-stack-showroom$ oc get llsd/llamastack-distribution -o yaml | yq .status.version
lastUpdated: "2026-04-11T21:50:01Z"
llamaStackServerVersion: 0.7.1+rhaiv.1
operatorVersion: '"0.4.0"'
```

VaishnaviHire:
Closing this as it is no longer needed

PR #288, opened 2026-04-24
feat: Rename Llama Stack to OGX

VaishnaviHire:
## Summary

- Introduce spec files for the OGX (Open GenAI Stack) operator rename
- Replace `LlamaStackDistribution` (`llamastack.io/v1alpha1`) with `OGXServer` (`ogx.io/v1beta1`) as a breaking change, folding in the expanded API surface from spec 002 (providers, resources, state storage, networking, workload, overrideConfig)
- No conversion webhooks — clean break with annotation-driven migration for PVC and networking adoption

## Key Decisions

1. **Breaking change, no webhooks**: The old CRD is replaced entirely. Users manually create new OGXServer CRs and migrate configuration.

2. **Spec 002 types folded in**: The new `OGXServer` CRD includes the expanded API surface from spec 002. Config generation logic is deferred to a follow-up PR.

3. **3-PR implementation strategy**:
   - **PR 1**: API types + generated artifacts (CRD, deepcopy, samples, CI)
   - **PR 2**: Controller + adoption logic + tests + migration guide
   - **PR 3**: Config generation (spec 002 follow-up)

4. **Upstream contracts out of scope**: Runtime contracts (`LLAMA_STACK_CONFIG`, `/etc/llama-stack/config.yaml`, etc.) are being updated upstream — deferred to a follow-up.

## Upgrade Path

```text
Step 1: Remove old LLS operator
Step 2: Delete orphaned stateless resources (keep PVC + optionally Service/Ingress)
Step 3: Install new OGX operator
Step 4: Create OGXServer CR with ogx.io/adopt-storage annotation
Step 5: (Optional) Add ogx.io/adopt-networking annotation
Step 6: Delete old CRD

### CODE REVIEW

File: specs/003-ogx-rename/spec.md [hunk 1]
```diff
@@ -0,0 +1,160 @@
...
+- **FR-013**: The operator MUST suppress creation of a new PVC from the kustomize manifest pipeline when the `ogx.io/adopt-storage` annotation is present, to avoid creating a duplicate empty PVC.
+- **FR-014**: The adoption logic MUST be idempotent: if the operator restarts or re-reconciles, it MUST detect that the PVC is already adopted (via owner reference check) and not re-run destructive steps such as scaling down the old deployment.
+- **FR-015**: The operator MUST support an `ogx.io/adopt-networking` annotation on the OGXServer custom resource whose value is the name of a legacy LlamaStackDistribution resource. When present, the operator MUST adopt the legacy Service (`{value}-service`) and Ingress (`{value}-ingress`), transfer their owner references to the new CR, and update the adopted Service's selectors to match the new pod labels. Since the OGXServer CR is expected to use the same name as the old LLSD, the adopted resource names match the reconciler's expected naming convention and no duplicate resources are created.
```
eoinfennessy: > Since the OGXServer CR is expected to use the same name as the old LLSD, the adopted resource names match the reconciler's expected naming convention and no duplicate resources are created.

There's no hard requirement that new OGXServer CRs should match the existing LLSD CR name. We should specify what happens when the names differ. In this case the controller should probably adopt the old Service and Ingress in addition to creating a new Service and Ingress.

specs/003-ogx-rename/spec.md
```diff
@@ -0,0 +1,160 @@
...
+- **FR-016**: Both adopted PVC and adopted Ingress MUST be reproducible from the annotation value plus the OGXServer CR spec. If an adopted resource is accidentally deleted, the operator MUST be able to recreate or re-adopt it using the same annotation.
```
eoinfennessy: We should mention Service CRs here too.

specs/003-ogx-rename/spec.md
```diff
@@ -0,0 +1,160 @@
...
+- **FR-002**: Upstream runtime contracts (server startup module paths, `LLAMA_STACK_CONFIG`, `/etc/llama-stack/config.yaml`, `/.llama`, etc.) are currently being updated by the upstream project. Preserving or renaming these is **out of scope** for the initial PRs. Handle in a follow-up once upstream stabilizes.
+- **FR-003**: The custom resource kind MUST be renamed to a name that reflects that the resource encompasses the full server deployment (not just a distribution selection).
+- **FR-004**: The custom resource status field that reports the running server version MUST be renamed to match the new naming, and the rename MUST be documented as a breaking change for status consumers.
+- **FR-005**: The operator's labels, annotations, leader-election identifier, operator-namespace name, and sample custom resources MUST all be updated to the new naming.
+- **FR-006**: Container image references, registry locations, and Git/documentation URLs that still reference the legacy organization MAY remain temporarily, but the rename plan MUST enumerate them so they can be migrated on a later schedule.
+
+#### Annotation-driven migration
```
eoinfennessy, line 105: We should add an FR for validating that adoption annotation values are non-empty and valid kubernetes resource names

specs/003-ogx-rename/research.md
```diff
@@ -0,0 +1,207 @@
...
+
+### D-006: Label and Annotation Values
+
+| Old Value | New Value | Notes |
+|-----------|-----------|-------|
+| `app: llama-stack` | `app: ogx` | Default label on workloads. New deployments use `app: ogx`. Old deployments are not re-parented (scaled to zero instead); see D-022. |
+| `app.kubernetes.io/managed-by: llama-stack-operator` | `app.kubernetes.io/managed-by: ogx-operator` | Used in cache selector, managed ConfigMap labels, Ingress labels. |
+| `app.kubernetes.io/part-of: llama-stack` | `app.kubernetes.io/part-of: ogx` | Kustomize-injected label on manifests. |
+| `app.kubernetes.io/name: llama-stack-k8s-operator` | `app.kubernetes.io/name: ogx-k8s-operator` | Kustomize default label. |
+| `llamastack.io/watch: "true"` | `ogx.io/watch: "true"` | Watch label key for ConfigMap cache filter. During transition, accept both. |
```
eoinfennessy, line 52: While I think of it, we should specify in our docs that all referenced CMs _must_have this label so that we can implement caching without requiring polling resources at regular intervals or using a direct (uncached) client. We should fail reconciliation and update CR status accordingly if any referenced CM does not have this label.

This is kind of unrelated to this spec.

eoinfennessy (commented):
A few small comments.

specs/003-ogx-rename/research.md
VaishnaviHire: Added a note about this, for follow-up work

VaishnaviHire:
@eoinfennessy Addressed all the comments. Please take a look

rhdedgar (approved):
It looked good here, +1

rhdedgar:
There's just one pre-commit hook to address around whitespace that I noticed.

eoinfennessy (approved):

specs/003-ogx-rename/spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+- **FR-007**: The operator MUST validate that the values of the `ogx.io/adopt-storage` and `ogx.io/adopt-networking` annotations, when present, are non-empty and conform to valid Kubernetes resource names (RFC 1123 DNS label: lowercase alphanumeric or `-`, at most 63 characters, starting and ending with an alphanumeric character). If validation fails, the operator MUST reject the CR by setting a status condition (e.g., `AdoptionConfigInvalid`) with a descriptive message and MUST NOT proceed with adoption or reconciliation of adopted resources.
```
mfleader: FR-007 is ambiguous about rejection scope. "MUST NOT proceed with adoption or reconciliation of adopted resources" could mean the entire reconcile halts or only adoption steps are skipped. If we skip adoption but still run normal reconciliation, we create a fresh PVC alongside the legacy one. If we halt everything, a typo in the annotation blocks the workload from deploying. Pick one and state it explicitly.

specs/003-ogx-rename/spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+- **FR-010**: The new operator MUST support an `ogx.io/adopt-storage` annotation on the OGXServer custom resource whose value is the name of a legacy LlamaStackDistribution resource in the same namespace. When present, the operator MUST adopt the legacy PVC (`{value}-pvc`) instead of creating a new one.
+- **FR-011**: When the `ogx.io/adopt-storage` annotation is present, the operator MUST transfer the adopted PVC's owner reference from the legacy custom resource to the new OGXServer custom resource, ensuring the PVC's lifecycle is tied to the new resource.
```
mfleader: FR-011 requires transferring PVC ownerReferences, which is an update operation. Current RBAC at `controllers/kubebuilder_rbac.go:26` only grants `get;list;watch;create` for PVCs. We need `update`/`patch` added, and those permissions should be marked transitional per FR-050.

specs/003-ogx-rename/plan.md
```diff
@@ -0,0 +1,207 @@
...
+
+- ✅ §1.1 Namespace-scoped — Adoption operates within the CR's namespace; no cluster-admin required.
+- ✅ §1.2 Idempotent reconciliation — Adoption checks ownerRefs before acting.
+- ✅ §1.3 Owner references — Adopted PVC/Service/Ingress get ownerRefs pointing to the new CR.
+- ✅ §2.1 Kubebuilder validation — New CRD uses CEL validation markers.
+- ✅ §3 Status has Phase + Conditions — `StorageAdopted` and `NetworkingAdopted` condition types.
+- ✅ §4.1 Error wrapping — All new code wraps errors with `%w` and resource identifiers.
+- ✅ §5.1 Logger in context — Adoption uses context logger with migration-specific fields.
+- ✅ §6.1 Table-driven tests — All adoption test cases use table-driven pattern.
+- ✅ §11 Feature flags — Adoption is opt-in via CR annotations; no global flag needed.
```
mfleader, line 35: 1. "No global flag needed" contradicts US2-AS2 (spec line 43), which requires a `--disable-legacy-check` flag. Annotations handle per-CR adoption at reconcile time, but US2-AS1's startup CRD check runs at boot. Different code paths.
  2. Constitution section 11 requires feature flags to use boolean fields in the reconciler. This check should confirm that pattern applies to `--disable-legacy-check`, not dismiss the flag.

specs/003-ogx-rename/tasks.md
```diff
@@ -0,0 +1,405 @@
...
+
+- [ ] T050 Update `pkg/deploy/kustomizer.go`: rename type refs, `FieldOwner` to `"ogx-operator"`, NS fallback to `"ogx-k8s-operator-system"`
+- [ ] T051 Update `pkg/deploy/deploy.go`: rename type refs and `FieldOwner`
+- [ ] T052 Update `pkg/deploy/utils.go` and `pkg/deploy/networkpolicy.go`: rename type refs
+- [ ] T053 Update `pkg/deploy/plugins/networkpolicy_transformer.go` and `plugins/field_mutator.go`: rename type refs
+- [ ] T054 Update `pkg/cluster/cluster.go`: managed-by label check from `"llama-stack-operator"` to `"ogx-operator"`
+
+### Main entrypoint
+
+- [ ] T055 Update `main.go`: rename import alias to `ogxiov1beta1`, update `LeaderElectionID` to `"54e06e98.ogx.io"`, update cache selector managed-by label, update reconciler call
```
mfleader, line 158: Related to the plan's missing startup legacy check and disable flag. T055 updates `main.go` but has no subtasks for the startup legacy CRD check or the `--disable-legacy-check` flag. Both need implementation and test tasks.

specs/003-ogx-rename/spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+A cautious administrator wants visibility into what the migration did and the ability to verify before committing to final cleanup of legacy resources.
+
+**Why this priority**: Migrations that are opaque or irreversible feel risky. Giving administrators clear audit trails and a deferred cleanup step increases confidence in the process.
+
+**Independent Test**: After an automated migration, inspect the cluster and verify every adopted resource carries an annotation pointing back to its legacy origin, and verify the legacy resources still exist (not auto-deleted) so the administrator can manually decide when to remove them.
+
+**Acceptance Scenarios**:
+
+1. **Given** adoption has completed, **When** the administrator lists new-kind resources, **Then** every migrated resource carries a machine-readable annotation identifying its legacy origin and the migration timestamp.
```
mfleader, line 72: This requires annotations on the adopted child resources themselves (`ogx.io/adopted-from`, `ogx.io/adopted-at`). FR-017 only covers conditions and events on the CR. No FR implements the child resource annotations, so someone working from the FRs alone would miss this.

specs/003-ogx-rename/spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+**Why this priority**: New users should not pay a cost for a migration they do not need. The operator must start up quickly and not create spurious warnings or perform unnecessary cluster-wide list operations.
+
+**Independent Test**: On a cluster with no legacy custom resource definition and no legacy custom resources, install the new operator and verify startup completes cleanly with no migration activity logged and no errors related to missing legacy resources.
+
+**Acceptance Scenarios**:
+
+1. **Given** a cluster with no legacy custom resource definition, **When** the new operator starts, **Then** the startup log confirms no legacy resources were found and no errors are produced.
+2. **Given** a cluster with no legacy custom resource definition, **When** the administrator disables legacy adoption via a flag, **Then** the operator skips the legacy check entirely and starts faster.
+3. **Given** a fresh install, **When** the administrator creates a new-kind custom resource using the sample manifest, **Then** the workload comes up successfully with no reference to the legacy naming anywhere.
```
mfleader: This AS contradicts FR-002, which explicitly preserves upstream contracts like `LLAMA_STACK_CONFIG` and `/etc/llama-stack/`. Qualify it to exclude upstream runtime contracts.

specs/003-ogx-rename/spec.md
mfleader: Kubernetes refuses a second controller ownerRef on the same resource. The old ownerRef must be removed before adding the new one, or the API call fails at runtime. I think we should state that explicitly.

mfleader (changes requested):

specs/003-ogx-rename/spec.md
```diff
@@ -0,0 +1,163 @@
...
+
+- **FR-001**: The operator's identity (name, repository, module path, container image, labels, API group, resource kind, and short name) MUST be migrated from all forms of "llama", "LlamaStack", and "llama-stack" to the corresponding "ogx" (Open GenAI Stack) naming, with the sole exceptions listed in FR-002.
+- **FR-002**: Upstream runtime contracts (server startup module paths, `LLAMA_STACK_CONFIG`, `/etc/llama-stack/config.yaml`, `/.llama`, etc.) are currently being updated by the upstream project. Preserving or renaming these is **out of scope** for the initial PRs. Handle in a follow-up once upstream stabilizes.
+- **FR-003**: The custom resource kind MUST be renamed to a name that reflects that the resource encompasses the full server deployment (not just a distribution selection).
+- **FR-004**: The custom resource status field that reports the running server version MUST be renamed to match the new naming, and the rename MUST be documented as a breaking change for status consumers.
+- **FR-005**: The operator's labels, annotations, leader-election identifier, operator-namespace name, and sample custom resources MUST all be updated to the new naming.
```
mfleader, line 102: FR-005 says "update to new naming" but never lists the actual replacement values. Add a rename mapping table covering: `llamastack.io/watch` label, `app.kubernetes.io/managed-by` value, leader-election ID (`54e06e98.llamastack.io`), operator namespace (`llama-stack-k8s-operator-system`), and SSA field owner (`client.FieldOwner("llama-stack-operator")` at `pkg/deploy/deploy.go:46`). I think missing any one breaks runtime behavior or causes field ownership conflicts.

mfleader (changes requested):

specs/003-ogx-rename/spec.md
VaishnaviHire: These are included in [tasks.md](https://github.com/ogx-ai/ogx-k8s-operator/pull/288/changes#diff-f2628e3f65ce809f4b8eb8451c46ddf5701b5c575aec7f747d6e546d4f091826R69)

specs/003-ogx-rename/spec.md
VaishnaviHire: Kubernetes does allow multiple ownerReferences on a resource. What it restricts is multiple entries with `controller: true`. In practice, the adoption logic will replace the old ownerRef (pointing to the deleted/orphaned legacy CR) with the new one (pointing to the OGXServer CR). If the old operator was removed and the LLSD CR was deleted with --cascade=orphan, the old ownerRef may already be stale. I can add explicit language to FR-011 stating the operator replaces (removes old + adds new) the ownerRef rather than appending.

specs/003-ogx-rename/spec.md
VaishnaviHire: I'll clarify FR-007 to state explicitly: `the operator MUST skip adoption and proceed with normal reconciliation using default resource naming`

specs/003-ogx-rename/spec.md
VaishnaviHire: This is already tracked as task T070 in tasks.md, which adds update permissions for PVCs

specs/003-ogx-rename/plan.md
VaishnaviHire: The constitution check is specifically about §11 (feature flags for adoption behavior), which is correctly opt-in via CR annotations, no global flag needed for that. US2-AS2 describes a capability where an admin can skip the legacy CRD check at startup. This is an acceptance scenario, not an FR, there's no FR mandating a --disable-legacy-check flag. If we decide to add one, it would be a simple operator argument, but it's not required by any FR and the startup CRD check is a lightweight operation (single API call to check if a CRD exists). We can track adding this flag as a follow-up.

specs/003-ogx-rename/tasks.md
VaishnaviHire: As noted in the previous comment, there's no FR requiring a `--disable-legacy-check` flag or a startup CRD scan.

specs/003-ogx-rename/plan.md
mfleader: That makes sense we can defer the feature flag for adoption behavior. I think that means we should delete the US2-AS2 for the flag then.

VaishnaviHire:
@mfleader Is this good to merge?

specs/003-ogx-rename/plan.md
VaishnaviHire: Done

PR #253, opened 2026-02-23
feat: Introduce v1alpha2 version of LlamaStackDistribution CRD

VaishnaviHire:
This PR introduces the `v1alpha2` API version for the `LlamaStackDistribution` CRD, enabling declarative, Kubernetes-native configuration of LlamaStack servers. Instead of requiring users to manually craft and supply a `config.yaml` via ConfigMap (as in v1alpha1), the operator now generates the server configuration automatically from structured CR fields (`providers`, `resources`, `storage`, `networking`). Both API versions are served concurrently with full conversion webhook support.

## v1alpha2 Example

The v1alpha2 API replaces environment-variable-driven configuration with structured, declarative fields. All provider fields use typed `[]ProviderConfig` slices with CEL validation.

**Basic : single Ollama provider:**

```yaml
apiVersion: llamastack.io/v1alpha2
kind: LlamaStackDistribution
metadata:
  name: llamastackdistribution-v1alpha2-sample
spec:
  distribution:
    name: starter
  providers:
    inference:
      - provider: ollama
        endpoint: http://ollama-server-service.ollama-dist.svc.cluster.local:11434/v1
  resources:
    models:
      - name: "llama3.2:1b"
  networking:
    port: 8321
  workload:
    replicas: 1
```

**Advanced : vLLM with secret refs, PostgreSQL storage, pgvector:**

```yaml
apiVersion: llamastack.io/v1alpha2
kind: LlamaStackDistribution
metadata:
  name: llamastack-vllm-pg
spec:
  distribution:
    name: starter
  providers:
    inference:
      - provider: vllm
        endpoint: http://vllm-service.vllm.svc.cluster.local:8000/v1
        secretRefs:
          api_key:
            name: vllm-creds
            key: token
    vectorIo:
      - provider: pgvector
        secretRefs:
          host:
            name: pg-credentials
            key: host
        settings:
          port: 5432
          db: llamastack
  resources:
    models:
      - name: llama3.2-8b
  storage:
    kv:
      type: redis
      endpoint: redis://redis-service.redis.svc.cluster.local:6379
    sql:
      type: postgres
      connectionString:
        name: pg-credentials
        key: dsn
  disabled:
    - safety
  workload:
    replicas: 2
```

## Review Guide

This is a large PR (86 files, ~18k lines). The sections below group changes by area matching the commit structure. Each section is self-contained and can be reviewed independently.

### 1. v1alpha2 CRD Schema & Conversion (commit: `e7cc5c2`)

| File | What to review |
[...]

PR #289, opened 2026-04-27
feat(api): Add OGXServer CRD types and renamed operator references

VaishnaviHire:
- Introduce the `OGXServer CRD` under `ogx.io/v1beta1` with the full expanded API surface from spec 002 (distribution, providers, resources, state storage, network, workload, externalProviders, overrideConfig) plus adoption annotation helpers and CEL validations.
- Redesign `spec.network` to replace AllowedFromSpec and the ConfigMap-based enableNetworkPolicy feature flag with NetworkPolicySpec using native Kubernetes NetworkPolicyIngressRule/NetworkPolicyEgressRule types and a per-CR enabled toggle (hybrid approach).
- Rename the operator identity across go.mod, config scaffolding, RBAC, kustomize, CI workflows, Makefile, and all Go import paths from llama-stack to ogx. Legacy api/v1alpha1 is preserved for controller compatibility until PR2.


eoinfennessy:
The improvements from the reviews of #253 are not present in this PR.

For instance, the API types are still using the polymorphic raw JSON instead of provider lists, and most of the kubebuilder validations have been removed.

Can we point the agent to take all improvements from https://raw.githubusercontent.com/VaishnaviHire/llama-stack-k8s-operator/2de6d1ef2bb1981450d86bc34f3c52dacdc2ebd8/api/v1alpha2/llamastackdistribution_types.go into account in this PR?

If you want you can make me a collaborator on your fork and I can work on incorporating some of these changes, allowing you to focus on the config gen pipeline, controller, etc?

VaishnaviHire:
> https://raw.githubusercontent.com/VaishnaviHire/llama-stack-k8s-operator/2de6d1ef2bb1981450d86bc34f3c52dacdc2ebd8/api/v1alpha2/llamastackdistribution_types.go

Yeah I missed the updates here. Added you as collaborator. Thanks !

eoinfennessy:
@VaishnaviHire @rhdedgar @mfleader some of the changes I've made diverge from the speckit spec. Summarising these changes and why they are present below.

1. Move `spec.network.tls.caBundle` to `spec.caBundle`: The CA bundle is a trust concern, not a networking concern.
2. Remove `spec.network.tls.enabled` (use presence semantics instead): `spec.network.tls` only has one field, the secret ref, which is a required field. A TLS spec with explicit `enabled: false` plus a required secret ref is contradicting and useless state that should be unrepresentable.
3. Rename `spec.network.networkPolicy` to `spec.network.policy`: Avoid redundant stutter
4. Add `policyTypes` to `spec.network.policy`: follow NetworkPolicy spec and semantics. Docstrings clearly describe the behaviour.
5. Rename `spec.network.expose` to `spec.network.externalAccess`. The verb "expose" really only made sense in the original polymorphic plan where it could be a boolean.
6. Add explicit `spec.network.externalAccess.enabled` field. This is clearer than the presence semantics of adding an empty object, which was unusual and didn't make it clear what the behaviour of an empty object was.
7. Remove `spec.caBundle.configMapNamespace`: Require all ConfigMaps and Secrets to be in the same namespace as the OGXServer for multi-tenant deployments
8. Remove `ExternalProviderRef`, `ExternalProvidersSpec`, and the `ExternalProviders` field: the design for external providers is not yet finalised, and adding these prematurely forces us into a corner.
9. Add docstrings to all ConfigMap and Secret references stating that they must be labelled `ogx.io/watch: "true"`: this will allow us to avoid cache issues from the beginning
10. Remove CEL rule ensuring ID is specified if more than one provider of the same API type is present: This was unnecessary as far as I can tell. What we need to ensure is that provider IDs (derived or explicit) are globally unique - I.e. all providers have unique IDs, regardless of API type. The validating webhook now does this.
11. Remove telemetry provider: it doesn't exist
12. Add CEL validation enforcing that `ProviderConfig.Provider` must start with `remote::` or `inline::`, removing implicit normalisation behaviour: Previously, `remote::` was added to all providers that did not specify a prefix - e.g. `vllm` became `remote::vllm`. This felt like it could be lead to UX issues - "When should I add a prefix?", "Why doesn't the `builtin` responses provider work?", etc. Explicitly requiring it is a better UX IMO, and removes the need for custom normalization logic.

config/samples/example-with-network-config.yaml
```diff
@@ -1,50 +1,81 @@
...
   network:
-    exposeRoute: false
-    allowedFrom:
-      namespaces:
-        - nsA
-        - nsB
-      labels:
-        - myproject/lls-allowed
-        - team/authorized
+    networkPolicy:
```
mfleader: I believe `networkPolicy` should be `policy`. The json tag on NetworkSpec.Policy is "policy,omitempty" (ogxserver_types.go L332), so the API server silently prunes `networkPolicy`  as an unknown field. Users lose their network policy config with no error.

config/samples/example-with-network-config.yaml
```diff
@@ -1,50 +1,81 @@
...
   network:
-    # Expose the service externally via Ingress
-    exposeRoute: true
-    allowedFrom:
-      # Allow access from all namespaces
-      namespaces:
-        - "*"
+    externalAccess:
+      enabled: true
+    networkPolicy:
```
mfleader: I believe `networkPolicy` should be `policy`. The json tag on NetworkSpec.Policy is "policy,omitempty" (ogxserver_types.go L332), so the API server silently prunes `networkPolicy`  as an unknown field. Users lose their network policy config with no error.

config/samples/example-with-network-config.yaml
```diff
@@ -1,50 +1,81 @@
...
+    name: starter
+  workload:
+    overrides:
+      env:
+        - name: OLLAMA_INFERENCE_MODEL
+          value: 'llama3.2:1b'
+        - name: OLLAMA_URL
+          value: 'http://ollama-server-service.ollama-dist.svc.cluster.local:11434'
+  network:
+    networkPolicy:
```
mfleader: I believe `networkPolicy` should be `policy`. The json tag on NetworkSpec.Policy is "policy,omitempty" (ogxserver_types.go L332), so the API server silently prunes `networkPolicy`  as an unknown field. Users lose their network policy config with no error.

mfleader (changes requested):

config/samples/example-with-network-config.yaml
eoinfennessy: Done in [9207c7e](https://github.com/ogx-ai/ogx-k8s-operator/pull/289/commits/9207c7e97ba9865122c5ce9afeda0ecbe78c2005)

config/samples/example-with-network-config.yaml
eoinfennessy: Done in [9207c7e](https://github.com/ogx-ai/ogx-k8s-operator/pull/289/commits/9207c7e97ba9865122c5ce9afeda0ecbe78c2005)

config/samples/example-with-network-config.yaml
eoinfennessy: Done in [9207c7e](https://github.com/ogx-ai/ogx-k8s-operator/pull/289/commits/9207c7e97ba9865122c5ce9afeda0ecbe78c2005)

PR #290, opened 2026-04-27
(feat)controller: Add ogx controller (reconciler rename)

VaishnaviHire:
Pre-req: Review and merge #289 
 
## Summary

- Renames `LlamaStackDistributionReconciler` → `OGXServerReconciler` and rewires
  the entire controller, `pkg/deploy`, and `pkg/cluster` to the `ogx.io/v1beta1`
  `OGXServer` CRD types introduced in the prior PR.
- Removes the global ConfigMap-based feature flag system (`pkg/featureflags`);
  NetworkPolicy enablement is now a per-CR toggle via
  `spec.network.networkPolicy.enabled` (defaults to true).
- Updates all kustomize manifests, labels (`ogx-operator`, `ogx`, `ogx.io/watch`),
  and the leader election ID to the new `ogx.io` identity.
- Migrates all controller and package tests to v1beta1 types and the new
  `OGXServerBuilder` test helper.

## Follow-up PRs 

- adoption logic + adoption tests
- remaining tests + docs + verification

## Test plan

- [ ] `go build ./...` passes
- [ ] `go vet ./...` passes
- [ ] `go test -c ./controllers/...` compiles
- [ ] `go test -c ./pkg/...` compiles
- [ ] Pre-commit hooks (linters, error message format) pass
- [ ] Envtest suite passes with kubebuilder assets
- [ ] Verify no remaining references to `v1alpha1`, `LlamaStackDistribution`,
      `llama-stack-operator`, or `llamastack.io` in `.go`/`.yaml` files

PR #292, opened 2026-05-05
feat(controller): add legacy resource adoption for OGXServer

VaishnaviHire:
**Pre-req:** Follow-up to #290 . That PR should be reviewed and merged first

This PR implements PR 2b of 003-ogx-rename: annotation-driven adoption of legacy LLSD resources into OGXServer, plus adoption-focused test coverage.

It adds controller logic to:

- adopt legacy storage (PVC) and networking (Service/Ingress) resources
- transfer ownership safely to the new OGXServer
- update adoption status conditions
- handle adoption-specific requeue flow
- clean up adopted networking when adoption annotations are removed
- Note invalid adoption annotations do not disrupt normal reconcile behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -276,7 +276,8 @@ docker-buildx: image-buildx ## Deprecated: use image-buildx instead
 build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
 	mkdir -p release
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default > release/operator.yaml
+	$(KUSTOMIZE) build config/overlays/cert-manager > release/operator.yaml
+	$(KUSTOMIZE) build config/overlays/openshift > release/operator-openshift.yaml
 
 .PHONY: image
 image: image-build image-push ## Build and push image with the manager.
@@ -296,13 +297,22 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: deploy
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+deploy: manifests kustomize ## Deploy controller to the K8s cluster (cert-manager overlay, vanilla K8s).
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/overlays/cert-manager | kubectl apply -f -
+
+.PHONY: deploy-openshift
+deploy-openshift: manifests kustomize ## Deploy controller to an OpenShift cluster (service-serving-cert-signer).
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/overlays/openshift | oc apply -f -
 
 .PHONY: undeploy
-undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+	$(KUSTOMIZE) build config/overlays/cert-manager | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+
+.PHONY: undeploy-openshift
+undeploy-openshift: kustomize ## Undeploy controller from an OpenShift cluster. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+	$(KUSTOMIZE) build config/overlays/openshift | oc delete --ignore-not-found=$(ignore-not-found) -f -
 
 ##@ Dependencies
 
--- a/api/v1beta1/ogxserver_types.go
+++ b/api/v1beta1/ogxserver_types.go
@@ -50,7 +50,10 @@ const (
 	AdoptStorageAnnotation = "ogx.io/adopt-storage"
 	// AdoptNetworkingAnnotation triggers Service/Ingress adoption from a legacy LlamaStackDistribution.
 	AdoptNetworkingAnnotation = "ogx.io/adopt-networking"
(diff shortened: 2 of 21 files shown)


eoinfennessy:
Thinking about this more, it's actually requirement that the new CR name must not equal the old CR name because this would cause more name conflicts (same Deployment name, etc.). This simplifies our logic because we now only have to account for resources of a different name.

I think we should add webhook validation to ensure annotation values do not equal the CR name.

I considered making this semi-idempotent by recreating adopted resources if they are accidentally deleted or updated, but this creates more problems because we do not know the spec for sure. It's more honest to not try to reconcile adopted resources.

So the logic becomes:

## Validation

| Condition | Result |
|---|---|
| `legacyName == instanceName` | Reject — `AdoptionConfigInvalid` |
| Invalid DNS label | Reject — `AdoptionConfigInvalid` |

## Storage — `ogx.io/adopt-storage`

| Scenario | Behavior |
|---|---|
| Legacy PVC exists | Scale down legacy Deployment, transfer ownerRef |
| Legacy PVC not found | Skip, pipeline creates `{instanceName}-pvc` normally |
| Annotation removed after adoption | PVC retained |

## Networking — `ogx.io/adopt-networking`

| Scenario | Behavior |
|---|---|
| Legacy Service exists | Transfer ownerRef, update selectors |
| Legacy Service not found | Skip |
| Legacy Ingress exists | Transfer ownerRef |
| Legacy Ingress not found | Skip |
| Annotation removed after adoption | Delete adopted Service and/or Ingress |

VaishnaviHire:
@eoinfennessy 
You are right that `legacyName == instanceName` is fundamentally broken at the Deployment level (same name collision). Rejecting it upfront is cleaner.

I added webhook and controller-side validations and simplified cleanup


Other major change- 
- Added `overlays/openshift` since we now create webhooks.

VaishnaviHire:
Did manual testing - 

## Setup

```bash
export NS=test-lls
export LEGACY_NAME=llamastackdistribution-sample
```

## 1. Verify legacy resources exist

```bash
oc get pvc ${LEGACY_NAME}-pvc -n $NS
oc get svc ${LEGACY_NAME}-service -n $NS
oc get deployment ${LEGACY_NAME} -n $NS
```

## 2. Remove the old operator

```bash
make undeploy
```

## 3. Orphan legacy resources

```bash
oc delete llamastackdistribution $LEGACY_NAME -n $NS --cascade=orphan

# Delete orphaned stateless resources — keep PVC + Service for adoption
oc delete deployment $LEGACY_NAME -n $NS
oc delete serviceaccount ${LEGACY_NAME}-sa -n $NS --ignore-not-found
oc delete networkpolicy $LEGACY_NAME -n $NS --ignore-not-found
oc delete hpa $LEGACY_NAME -n $NS --ignore-not-found
oc delete pdb $LEGACY_NAME -n $NS --ignore-not-found

# Verify PVC and Service survived (no ownerRefs)
oc get pvc ${LEGACY_NAME}-pvc -n $NS -o jsonpath='{.metadata.ownerReferences}'
```

## 4. Deploy the new operator (OpenShift overlay)

```bash
make deploy-openshift IMG=<your-image>
```

## 5. Create OGXServer with adoption annotations

```yaml
apiVersion: ogx.io/v1beta1
kind: OGXServer
metadata:
  name: ogx-sample
  namespace: test-lls
  annotations:
    ogx.io/adopt-storage: llamastackdistribution-sample
    ogx.io/adopt-networking: llamastackdistribution-sample
spec:
  distribution:
    name: starter
  workload:
    replicas: 1
    workers: 2
    storage:
      size: 20Gi
      mountPath: /home/lls/.lls
    autoscaling:
      maxReplicas: 5
      minReplicas: 1
      targetCPUUtilizationPercentage: 75
      targetMemoryUtilizationPercentage: 70
    podDisruptionBudget:
      maxUnavailable: 1
    topologySpreadConstraints:
      - labelSelector:
          matchLabels:
            app.kubernetes.io/instance: ogx-sample
        maxSkew: 1
        topologyKey: topology.kubernetes.io/zone
        whenUnsatisfiable: ScheduleAnyway
    overrides:
      env:
        - name: OLLAMA_INFERENCE_MODEL
          value: "llama3.2:1b"
        - name: OLLAMA_URL
          value: "http://ollama-server-service.ollama-dist.svc.cluster.local:11434"
  network:
    externalAccess:
      enabled: true
```

## 6. Verify adoption

```bash
# PVC ownership transferred
oc get pvc ${LEGACY_NAME}-pvc -n $NS -o jsonpath='{.metadata.ownerReferences[0].kind}'
# expect: OGXServer

# Adoption annotations present
oc get pvc ${LEGACY_NAME}-pvc -n $NS -o jsonpath='{.metadata.annotations.ogx\.io/adopted-from}'
# expect: llamastackdistribution-sample

# Service selectors updated to new pods
oc get svc ${LEGACY_NAME}-service -n $NS -o jsonpath='{.spec.selector}'
# expect: {"app":"ogx","app.kubernetes.io/instance":"ogx-sample"}

# New service also exists (different name from adopted one)
oc get svc ogx-sample-service -n $NS

# Status conditions
oc get ogxserver ogx-sample -n $NS -o jsonpath='{.status.conditions}' | python3 -m json.tool
# expect: StorageAdopted=True, NetworkingAdopted=True

# Deployment running with adopted PVC mounted
oc get pods -l app.kubernetes.io/instance=ogx-sample -n $NS
```

eoinfennessy:
Looks good apart from one or two issues.

When the `ogx.io/adopt-storage` annotation is removed, a new PVC is created and used with the deployment. Is this the correct behaviour? I would expect the deployment to continue using the adopted PVC if it is available. Unlike networking CRs, I think we need to ensure the adopted PVC continues to be used after the annotation is removed. Probably the most robust way to do that is to check if there is a PVC in the same namespace whose controller reference is the OGXServer CR and contains the `adopted-from` label. If this exists we should use it instead of creating a new PVC.

We can remove the DELETE verb from RBAC for PVCs - it is unused.

Also, we can use labels instead of annotations for `adopted-from`. This allows us to filter at the list request instead of client side.

VaishnaviHire:
> We can remove the DELETE verb from RBAC for PVCs - it is unused.

delete verb is needed when CR is deleted. The current LlamaStackDistribution CR manages the PVCs and deleting CRs cleans up data as well. Same behavior is reflected in OGServer. I believe CR deletion is a breaking operation for user.

VaishnaviHire:
@eoinfennessy Addressed comments. This is ready for review again

eoinfennessy:
Added a few changes here: https://github.com/VaishnaviHire/llama-stack-k8s-operator/pull/4

eoinfennessy (approved):
Looks good!

PR #295, opened 2026-05-08
Add config generation pipeline

VaishnaviHire:
This PR implements and documents the OGX runtime config generation pipeline driven by OGXServer CR fields.

Runtime config generation
Generates config.yaml from declarative fields:
```
spec.providers
spec.resources
spec.storage
spec.disabledAPIs
```

Uses OCI label `com.ogx.distribution.configs` as primary base config source, with embedded config fallback.
Creates immutable generated ConfigMaps (<name>-config-<hash>) and mounts generated config.yaml to the server pod.
Preserves precedence: spec.overrideConfig continues to override declarative generation.


VaishnaviHire:
@eoinfennessy This PR is ready to be reviewed

eoinfennessy:
@mfleader @rhdedgar you might also want to review this

eoinfennessy:
The OCI label config resolution path doesn't work - it currently just reads a list of filenames instead of the config file itself.

See here for the label generation: https://raw.githubusercontent.com/ogx-ai/ogx/refs/heads/main/scripts/generate-config-labels.sh 

The OCI label config resolution path should...
1. Check for the default config filename (`com.ogx.distribution.default-config`), then
2. decode `com.ogx.config.<default-config-filename>`.

eoinfennessy:
I think the embedded configs could be removed. We will run into issues where the embedded configs become stale and we could potentially break upgrades if we need to change them.

Instead, we could add an optional ConfigMap ref for a base config to the CR. If this is set, use it, otherwise use OCI labels. WDYT?

eoinfennessy:
@VaishnaviHire, I pushed a command adding a CLI that runs the entire config generation pipeline. This allows us to generate config without Kubernetes, and with the addition of https://github.com/ogx-ai/ogx/pull/5983 we can now pipe the output to the ogx CLI to ensure it is valid.

The pipeline runs from an input OGXServer yaml to the config.yaml + metadata. A path to an override base config.yaml can be provided to bypass OCI / embedded resolution. Validation of the CR can also be enabled.

I also setup a test scaffold that allows us to provide a `cr.yaml` and optional `base.yaml` and assert it structurally matches `want-config.yaml`. This has already picked up two issues in the basic happy path test (see the TODOs in the `want-config.yaml` for missing `distro_name` and incorrect nesting of `models`. We should add more test cases for this, parts of our base config here could help with these: https://github.com/opendatahub-io/ogx-distribution/blob/main/distribution/config.yaml

We should also test with the ogx server to ensure the config is actually valid and that it is reflected in the output of `--dry-run`.

eoinfennessy:
A few issues with dropped base config fields:

- `server.auth` is always dropped from the base config. `buildServerSection` only copies `port`, not `auth` or other server fields.
- `registered_resources`, `vector_stores`, `telemetry` dropped from base config. `BaseConfig` doesn't have fields for these, so they're lost during parse.

VaishnaviHire:
@eoinfennessy I addressed the comments regarding baseConfig
- add optional `spec.baseConfig`
- resolve OCI base config via `com.ogx.distribution.default-config` + `com.ogx.config.<filename>`
- remove embedded operator-shipped base configs
- keep `spec.baseConfig` as generation input only; the mounted runtime config still comes from `overrideConfig` or the generated ConfigMap

Open question
1. Midstream/downstream distributions still use
- `registered_resources.vector_dbs`
- possibly `registered_resources.tool_groups`
- storage merge behavior when only part of `spec.storage` is set.
Should we update the distribution config ?

pkg/config/resource.go
```diff
@@ -0,0 +1,64 @@
...
+	// Expand models
+	for _, m := range resources.Models {
+		providerID := m.Provider
+		if providerID == "" {
+			// Default to first inference provider
+			inferenceProviders := providers["inference"]
+			if len(inferenceProviders) == 0 {
+				return nil, fmt.Errorf("failed to assign provider for model %q: no explicit provider and no inference providers are configured", m.Name)
+			}
+			providerID = inferenceProviders[0].ProviderID
```
rhdedgar: I'm thinking this should be updated to be deterministic. Like by sorting the inference providers slice by `ProviderID` before returning from `ExpandProviders()`, or require an explicit provider on models when multiple inference providers exist.

In theory two reconciliation loops could assign the same model to different providers.

pkg/config/oci_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+			continue
+		}
+		resp, err := ociHTTPClient.Do(req)
+		if err != nil {
+			continue
+		}
+		_ = resp.Body.Close()
+		return fmt.Sprintf("%s://%s", scheme, registry), nil
+	}
+	return "", errors.New("failed to connect on https or http")
```
rhdedgar, line 169: Currently, `detectRegistryScheme()` will silently fall back from HTTPS to HTTP when HTTPS fails. I think this would be a concern in FIPS environments. Maybe add a warning-level log when falling back to HTTP so there is some audit trail of the event.

controllers/resource_helper.go
```diff
@@ -259,19 +291,33 @@ func configureContainerEnvironment(ctx context.Context, r *OGXServerReconciler,
...
+	// Inject pre-computed secret env vars for provider/storage references.
+	// These are computed once in buildManifestContext to avoid redundant tree walks.
+	if len(secretEnvVars) > 0 {
+		container.Env = append(container.Env, secretEnvVars...)
+	}
+
 	// Finally, add the user provided env vars
 	if instance.Spec.Workload != nil && instance.Spec.Workload.Overrides != nil {
 		container.Env = append(container.Env, instance.Spec.Workload.Overrides.Env...)
 	}
```
rhdedgar, line 321: Operator-set env vars (`OGX_WORKERS`, `OGX_PORT`, `OGX_CONFIG`) are appended first, then user overrides are appended after. If a user sets `OGX_WORKERS=2` in their overrides, they get a duplicate entry. Kubernetes takes the first occurrence, so the user's override is silently ignored.

Maybe we could deduplicate so that user overrides take priority, or reject user overrides that collide with operator-managed env vars via webhook validation.

pkg/config/resolver.go
```diff
@@ -0,0 +1,139 @@
...
+	return data, ok
+}
+
+func (c *ociConfigCache) set(key string, data []byte) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if len(c.store) >= maxOCICacheEntries {
+		c.store = make(map[string][]byte)
+	}
+	c.store[key] = data
```
rhdedgar: What do you think about using something like an LRU eviction strategy (evicting oldest N entries) or even using a time-based TTL per entry? 

At the moment, when the cache hits 64 entries, it clears the entire map. Under load with many OGXServer instances referencing different images, this could create a thundering-herd pattern where all configs get evicted simultaneously, and then all instances re-fetch on the next reconciliation.

rhdedgar (commented):
Hi, I added a few comments on things that we might want to revisit before merging, unless there are reasons to ignore or defer to follow-up PRs.

pkg/config/resource.go
eoinfennessy: Done in [84814f7](https://github.com/ogx-ai/ogx-k8s-operator/pull/295/commits/84814f73364efc575a37c5bfd42e84bdf5f5c232)

pkg/config/oci_fetcher.go
eoinfennessy: Done in [461f020](https://github.com/ogx-ai/ogx-k8s-operator/pull/295/commits/461f0207c4f2960414bc8484b80b28e65d13c1c7)

eoinfennessy: I'm honestly not sure we should be doing this fallback in the first place. @VaishnaviHire WDYT?

pkg/config/resolver.go
eoinfennessy: LRU eviction strategy added in [3a54286](https://github.com/ogx-ai/ogx-k8s-operator/pull/295/commits/3a5428694b43f889b77875a5b64fe53089792541)

pkg/config/oci_fetcher.go
VaishnaviHire: Yeah I think we can remove the fallback.

rhdedgar:
I updated `distributions.json` yesterday to match the new https://hub.docker.com/u/ogxai organization name, so there are a few conflicts to resolve. I also noticed the e2e-tests in https://github.com/ogx-ai/ogx-k8s-operator/actions/runs/27270861215/job/80539656337?pr=295 are hitting issues with server creation with errors like:

```
=== RUN   TestE2E/creation-deletion-starter/creation/should_create_OGXServer
    creation_test.go:34: Creating starter distribution with name: ogxserver-starter-sample
    creation_test.go:34: 
        Error Trace:	/home/runner/work/ogx-k8s-operator/ogx-k8s-operator/tests/e2e/creation_test.go:110
        	            				/home/runner/work/ogx-k8s-operator/ogx-k8s-operator/tests/e2e/creation_test.go:34
        Error:      	Received unexpected error:
        	            	context deadline exceeded
        Test:       	TestE2E/creation-deletion-starter/creation/should_create_OGXServer
```

I was able to replicate the error when running the tests locally against my OpenShift cluster as well.

VaishnaviHire:
@rhdedgar The failures in e2e tests were triggered by missing OCI labels in the `starter` distribution image. We now require distribution images to have an oci label with `config.yaml` for base config. 

I have added a temporary workaround to use user configmap overrides. Once the `starter` image is updated I will open a follow-up PR to remove the overrides

rhdedgar (approved):
I think this PR is good to go. Any nits or non-blocking issues can be addressed via a new Jira ticket in the same Epic.

PR #297, opened 2026-05-14
fix: Update startupScript with ogx

VaishnaviHire:
StartupScript that is used by custom user config still refers to llama-stack client. This commit fixes the references to point to ogx

### DIFF
--- a/controllers/resource_helper.go
+++ b/controllers/resource_helper.go
@@ -36,7 +36,7 @@ import (
 // Constants for validation limits.
 const (
 	// FSGroup is the filesystem group ID for the pod.
-	// This is the default group ID for the llama-stack server.
+	// This is the default group ID for the ogx server.
 	FSGroup = int64(1001)
 	// instanceLabelKey is the label we apply to all resources for per-instance targeting.
 	instanceLabelKey = "app.kubernetes.io/instance"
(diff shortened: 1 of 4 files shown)


controllers/resource_helper.go
```diff
@@ -93,19 +93,19 @@ except Exception as e:
     print(2)
 ")
 
-PORT=${LLS_PORT:-8321}
-WORKERS=${LLS_WORKERS:-1}
+PORT=${OGX_PORT:-8321}
```
skamenan7, line 96: the env var rename (LLS_PORT -> OGX_PORT, LLS_WORKERS -> OGX_WORKERS, LLAMA_STACK_CONFIG -> OGX_CONFIG) is a breaking change for users who set these in their CR's envVars. any existing CR with LLS_PORT: 8080 will silently fall back to the default 8321. does adding a fallback  (PORT=${OGX_PORT:-${LLS_PORT:-8321}}) or calling this out in the operator release notes help?

skamenan7 (approved):
LGTM, one comment. Thanks!

eoinfennessy (approved):

controllers/resource_helper.go
VaishnaviHire: Current upgrade is a breaking change. But I have called this out in migration guide https://github.com/ogx-ai/ogx-k8s-operator/pull/297/changes/c98ce6015dee3a2e9afe6f4656cfcf9f3a2b9ade#diff-c27572d4a0b169a6124ba513a3d6a6a9751cbd1203fe69600cf934534a4eb15bR26

nathan-weinberg (approved):

PR #302, opened 2026-06-02
feat: add CRD scaffolding for metrics and monitoring

rhdedgar:
This PR adds basic CRD support for Metrics endpoints.

Part of the Telemetry Epic RHAIENG-5155.

Closes RHAIENG-5404.


api/v1beta1/ogxserver_types.go
```diff
@@ -316,6 +318,21 @@ type NetworkSpec struct {
 	Policy *NetworkPolicySpec `json:"policy,omitempty"`
 }
 
+// MonitoringSpec configures Prometheus monitoring for this OGXServer instance.
+type MonitoringSpec struct {
+	// Enabled controls whether the operator creates monitoring resources
+	// (ServiceMonitor, PrometheusRule) for this server.
+	// +optional
+	// +kubebuilder:default:=false
+	Enabled bool `json:"enabled,omitempty"`
```
eoinfennessy: Should we default to true? If specifying `spec.monitoring: {...}` I would expect it to default to true - it follows the same pattern as other fields in the CRD. To disable monitoring users either explicitly set `enabled: false`, or omit `monitoring` from the spec.

Alternatively we could remove `enabled` from `monitoring`, and rely purely on presence semantics.

```suggestion
	// +kubebuilder:default:=true
	Enabled *bool `json:"enabled,omitempty"`
```

api/v1beta1/ogxserver_types.go
```diff
@@ -316,6 +318,21 @@ type NetworkSpec struct {
...
+	// MetricsPort is the port serving the /metrics endpoint.
```
eoinfennessy, line 329: nit: present tense "serving" suggests something is already serving these metrics and could be confusing.

```suggestion
	// MetricsPort defines the port on which the /metrics endpoint will be exposed.
```

api/v1beta1/ogxserver_types.go
rhdedgar: Sure! I've updated the necessary code to match.

VaishnaviHire (approved):
Thanks

VaishnaviHire:
@mergify rebase

PR #305, opened 2026-06-10
feat: support for PrometheusRules and ServiceMonitors permission

rhdedgar:
Adds RBAC rules which allow for eventual Metrics / Telemetry integration.

Builds on the `CRD` support from RHAIENG-5404, which is to be implemented in #302. The code from these two PRs will be used and finalized in future tickets & PRs. 

Closes RHAIENG-5405.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -44,6 +44,14 @@ rules:
   - update
   - watch
 - apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - get
+  - list
+  - watch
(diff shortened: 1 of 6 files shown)


VaishnaviHire:
Should this also include adding the ServiceMonitor ?

rhdedgar:
I should probably keep this PR scoped to just these permissions changes, and do a follow-up PR to handle the `ServiceMonitor` portions. I will have at least one more PR to address that and other related changes in `RHAIENG-5406`.

VaishnaviHire:
@mergify rebase

VaishnaviHire (approved):

PR #306, opened 2026-06-11
feat: add registryRefreshIntervalSeconds field to OGXServer CRD

nathan-weinberg:
## Summary

- Adds an optional `spec.registryRefreshIntervalSeconds` field (`*int32`, minimum 1) to the OGXServer CRD for configuring the model registry refresh interval
- Passes the value as `OGX_REGISTRY_REFRESH_INTERVAL_SECONDS` env var to the container; omitted when unset so the server's built-in default applies
- Includes unit tests for env var emission and schema validation tests for the minimum constraint

Ref: RHAIENG-5504

## Test plan

- [x] `make lint` passes
- [x] `make test` passes (unit + integration)
- [ ] Deploy operator with a CR setting `registryRefreshIntervalSeconds: 60` and verify the env var appears on the pod

### DIFF
--- a/api/v1beta1/ogxserver_cel_test.go
+++ b/api/v1beta1/ogxserver_cel_test.go
@@ -866,6 +866,56 @@ func TestCEL_WorkloadOverrides(t *testing.T) {
 	}
 }
 
+func TestCEL_RegistryRefreshIntervalSeconds(t *testing.T) {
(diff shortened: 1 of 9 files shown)


VaishnaviHire (approved):
Thanks for the change

PR #309, opened 2026-06-17
fix: Add ReplicaSet cache and RBAC access for generated config cleanup

VaishnaviHire:
Generated config cleanup lists ReplicaSets to preserve ConfigMaps still referenced by in-flight rollouts. Add ReplicaSet cache coverage and RBAC so that lookup works reliably under the controller-runtime cached client.

### DIFF
--- a/main.go
+++ b/main.go
@@ -114,7 +114,6 @@ func newCacheOptions() cache.Options {
 				}),
 			},
 			&appsv1.Deployment{}:                     managedByFilter,
-			&appsv1.ReplicaSet{}:                     managedByFilter,
 			&policyv1.PodDisruptionBudget{}:          managedByFilter,
 			&autoscalingv2.HorizontalPodAutoscaler{}: managedByFilter,
 			&corev1.Service{}:                        managedByFilter,
--- a/main.go
+++ b/main.go
@@ -114,6 +114,7 @@ func newCacheOptions() cache.Options {
 				}),
 			},
 			&appsv1.Deployment{}:                     managedByFilter,
+			&appsv1.ReplicaSet{}:                     managedByFilter,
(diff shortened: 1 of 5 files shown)


rhdedgar (approved):
Looks good to me.

main.go
```diff
@@ -114,6 +114,7 @@ func newCacheOptions() cache.Options {
 				}),
 			},
 			&appsv1.Deployment{}:                     managedByFilter,
+			&appsv1.ReplicaSet{}:                     managedByFilter,
```
mfleader: `managedByFilter` requires `app.kubernetes.io/managed-by: ogx-operator`, but ReplicaSets inherit labels from the Deployment's `.spec.template.metadata.labels` (which only has `app` and `instance`). The `managed-by` label is on `.metadata.labels` via kustomization `includeSelectors: false` and doesn't propagate. The cached client will silently return empty for the RS query in `referencedRuntimeConfigMapNames()`.

For example, a ReplicaSet created during rollout gets:

```yaml
metadata:
  labels:
    app: ogx
    app.kubernetes.io/instance: my-server
    pod-template-hash: abc123
    # no managed-by — cache filter excludes this RS
```

I believe the simplest fix might be to drop the `ByObject` entry for ReplicaSets and keep only the RBAC addition. The operator reads RS but doesn't own them, and `referencedRuntimeConfigMapNames()` already filters by instance label at query time.

mfleader (changes requested):

main.go
VaishnaviHire: Good catch. Fixed it. PTAL

VaishnaviHire:
@mergify rebase

mfleader (approved):

PR #320, opened 2026-06-30
fix: preserve image entrypoint when using overrideConfig

derekhiggins:
## Summary
- Stop replacing the container entrypoint with an inline startup script when overrideConfig is used
- Instead, set `RUN_CONFIG_PATH` and `OGX_CONFIG` env vars so the image's own entrypoint handles config selection
- Preserves OTEL auto-instrumentation and eliminates a divergent startup code path

## Details

When an OGXServer uses `overrideConfig`, the operator was replacing the container's entrypoint with a 45-line inline bash script that did version detection and direct uvicorn invocation. This bypassed the image's `entrypoint.sh`, losing OTEL wrapping and creating two divergent startup paths.

The fix lets the image entrypoint run normally and controls behavior via env vars:
- `RUN_CONFIG_PATH=/etc/ogx/config.yaml` - tells the entrypoint to use the mounted config
- `OGX_CONFIG=/etc/ogx/config.yaml` - used by `server.py:create_app()` factory

Both upstream and downstream container images already support `RUN_CONFIG_PATH` in their entrypoints.

Depends on [ogx#6221](https://github.com/ogx-ai/ogx/pull/6221) for `OGX_WORKERS` env var support (so worker count remains controllable via the CR spec).

Breaks backward compatibility with pre-0.3.0 (pre-rename) images - they didn't have the ogx module name and are already non-functional.

## Test plan
- [x] Unit tests pass (`make test`)
- [x] Verified on live cluster: with overrideConfig, PID 1 = `ogx run /etc/ogx/config.yaml`, OTEL active
- [x] Verified on live cluster: without overrideConfig, entrypoint uses baked-in config, no stale `OGX_CONFIG` env var

### DIFF
--- a/controllers/resource_helper.go
+++ b/controllers/resource_helper.go
@@ -300,22 +300,27 @@ func configureContainerEnvironment(
 		container.Env = append(container.Env, secretEnvVars...)
 	}
 
-	// Apply user-provided env vars, letting them override operator defaults.
-	if instance.Spec.Workload != nil && instance.Spec.Workload.Overrides != nil {
-		overrides := make(map[string]corev1.EnvVar, len(instance.Spec.Workload.Overrides.Env))
-		for _, e := range instance.Spec.Workload.Overrides.Env {
-			overrides[e.Name] = e
-		}
-		deduped := make([]corev1.EnvVar, 0, len(container.Env))
-		for _, e := range container.Env {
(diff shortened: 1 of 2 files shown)


derekhiggins:
Set to draft, pending the OGX change

rhdedgar (approved):
No concerns here, +1

PR #324, opened 2026-07-06
feat: add ServiceMonitor, PrometheusRule, and metrics NetworkPolicy

derekhiggins:
Add monitoring resource support to the operator:

- ServiceMonitor targeting metrics port 9464 with 60s scrape interval
- PrometheusRule with ogx:api_info recording rules for the Telemeter pipeline (inference, vector_io, responses, rag, agentic), aggregated via sum() to strip high-cardinality labels, within the ~10 timeseries budget
- NetworkPolicy ingress rule allowing Prometheus scraping from openshift-monitoring namespace on the metrics port
- Metrics endpoint env var injection (OGX_METRICS_ENDPOINT_ENABLED, OGX_METRICS_HOST, OGX_METRICS_PORT) and container port configuration
- Conditional gating on MonitoringCRDsAvailable for both resources

### DIFF
--- a/controllers/manifests/base/kustomization.yaml
+++ b/controllers/manifests/base/kustomization.yaml
@@ -10,6 +10,8 @@ resources:
 - hpa.yaml
 - pdb.yaml
 - rolebinding.yaml
+- prometheusrule.yaml
+- servicemonitor.yaml
 
 labels:
 - includeSelectors: false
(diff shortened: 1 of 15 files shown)


rhdedgar (approved):
Looks good! I think it addresses everything from `RHAIENG-5408`.

PR #326, opened 2026-07-08
test: add monitoring configuration and ServiceMonitor render tests

derekhiggins:
## Summary
- Adds `TestMonitoringConfiguration` controller integration test covering monitoring env vars, container ports, nil/enabled/disabled/custom-port cases
- Adds ServiceMonitor kustomize render test verifying name prefix and namespace application

## Test plan
- [x] `make lint` passes
- [ ] `make test` passes


rhdedgar (approved):
I'll have a couple follow-up items to submit after this merges but I think it's worth pushing through

PR #330, opened 2026-07-13
Scc updates tls

rhdedgar:
- Avoids running with `anyuid` permissions
- TLS is required for external access is enabled
- Fixes a few config and build errors that I encountered along the way. e.g. `llama`->`ogx` renaming, duplicate `boolPtr` func, etc.

Closes RHAIENG-5584, RHAIENG-5588, and RHOAIENG-76479.


VaishnaviHire:
Is the test failure expected?

rhdedgar:
Not this one. The others that were blocking other k8s-operator PRs were resolved, and now there is a new issue, possibly with ollama. I'm looking into it now.

rhdedgar:
Ok, so I found two sets of issues that were blocking this PR.

One was a new, unrelated issue with container image name formatting that I just fixed in #336

The other one is an upstream ogx server dependency, because the starter distribution image runs as root by default. I've created a PR here https://github.com/ogx-ai/ogx/pull/6319 to address that.

After that merges, then the next starter distribution container build should resolve this error.

VaishnaviHire:
@mergify rebase

VaishnaviHire:
@mergify rebase

VaishnaviHire:
@mergify rebase

config/samples/example-with-configmap.yaml
VaishnaviHire, line 50: nit: The sample OGServer CRs need update after the validation webhook updates. We need to update network field with -
```
network:
  externalAccess:
    enabled: true
    hostname: <hostname>
    tls:
      secretName: <tls-secret>
```

VaishnaviHire (commented):
Thanks for the updates. The improvements look good to me. 

Since this is a breaking change, we might want to update the sample CRs and documentation downstream

VaishnaviHire (approved):

PR #331, opened 2026-07-14
refactor: telemetry and related tests

rhdedgar:
Follow-up items for #326, adding `ServiceMonitor` labels, tests, and targeting the metrics port.

Closes RHAIENG-5409.

### DIFF
--- a/controllers/manifests/base/servicemonitor.yaml
+++ b/controllers/manifests/base/servicemonitor.yaml
@@ -2,6 +2,8 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: service-monitor
+  labels:
+    monitoring.opendatahub.io/scrape: "true"
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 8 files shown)


VaishnaviHire (approved):
Changes look good to me

controllers/manifests/base/servicemonitor.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: service-monitor
+  labels:
+    monitoring.opendatahub.io/scrape: "true"
```
derekhiggins, line 6: I mentioned this on the last PR but it got lost somewhere
don't we also need a label on the namespace 
`openshift.io/cluster-monitoring: "true"`
for Telemeter to get the metrics ?

rhdedgar: It would be required for Telemeter, but I think that's something the platform should apply instead of granting the operator those permissions to modify namespaces. I was considering it to be out of scope for the operator.

PR #353, opened 2026-08-20
feat: add Praxis-fronted internal-only deployment mode for OGXServer

nathan-weinberg:
## What

Adds an explicit **Praxis-fronted, internal-only deployment mode** to the OGX operator. OGX is an internal backend fronted by Praxis (the public entrypoint); in this mode the operator restricts OGX so it cannot be reached directly (bypassing Praxis auth/tenant normalization) or exposed externally.

This PR delivers **RHAIENG-6736** (internal-only deployment mode), which now subsumes **RHAIENG-6663** (Praxis-only NetworkPolicy ingress).

> **Depends on #355** (RHAIENG-7193), which introduces the `spec.praxisMode` API (`enabled`, `praxisSelector`, `migrationJob`) and the mutating webhook defaulter. This PR is stacked on that branch and only wires the operator to consume it. Retarget to `main` once #355 merges.

## Mode selection (`spec.praxisMode.enabled`, upgrade-safe)

Tri-state, resolved per-CR — no reconciliation state is derived from `status`:

- **`true`** — Praxis-fronted (internal-only).
- **`false`** — legacy behavior.
- **unset** — new CRs get `enabled` defaulted to `true` on create by #355's mutating webhook (greenfield → Praxis mode); CRs predating the webhook keep an unset value and are treated as **legacy**, so an operator upgrade never silently cuts off co-located workloads. Because defaulting is per-CR at create time, OGXServers migrate independently.

## Changes

- **Consume #355's API** — dropped this branch's own `spec.praxisMode *bool` field, its mutating defaulter, and the operator-wide `praxis-peer` ConfigMap mechanism in favor of #355's `spec.praxisMode` struct.
- **Per-CR Praxis peer** — the NetworkPolicy Praxis ingress peer is built from **`spec.praxisMode.praxisSelector`** (namespace + pod selector), so different OGXServers can front different Praxis instances (addresses @VaishnaviHire and @eoinfennessy's per-CR / multiple-Praxis-instances point). Falls back to `app: payload-processing` in `openshift-ingress` when unset; never emits an allow-all peer (the selector is CEL-validated non-empty in #355).
- **Praxis mode NetworkPolicy** — ingress on `8321` admits only the Praxis peer + the operator namespace (control-plane status polling). The broad same-namespace peer and OpenShift-router peer are removed. `spec.network.policy.ingress` rules are **additive** on top of the mandatory lock-down. `policy.enabled: false` still fully disables the policy.
- **Enforced internal-only external access (Praxis mode)** — the operator never creates an Ingress and removes any it previously created; `status.externalURL` is cleared while `status.serviceURL` keeps the internal cluster DNS endpoint. `spec.network.externalAccess.enabled: true` is not honored and surfaces an admission **warning** (not a rejection).
- **Responses API** — disabled in the generated config in Praxis mode (served by Praxis); applied internally, does not mutate `spec.disabledAPIs`.
- **Legacy mode** — preserves the pre-Praxis peers (same-namespace + OpenShift router) and `externalAccess`-driven Ingress (RHAIENG-6736 AC#8: router peers apply only in legacy mode).

Out of scope here (tracked by RHAIENG-7193/#355): wiring the DB **migration Job** (`spec.praxisMode.migrationJob`) and the target-PostgreSQL instance.

## Testing

- Unit / envtest (all non-e2e packages pass): `resolvePraxisMode` and `BuildPraxisPeer` (praxisSelector→peer mapping + default fallback, never allow-all); NetworkPolicy contains only Praxis + operator peers in Praxis mode and same-namespace + router in legacy; user rules merge additively; no Ingress created even with `externalAccess.enabled: true`; previously-owned Ingress deleted; `status.serviceURL` populated / `status.externalURL` empty; webhook warns (not errors) on `externalAccess.enabled: true`; Responses API disabled in generated config.
- CI-runnable e2e negative suite asserts on-cluster NetworkPolicy shape (now driven by per-CR `praxisSelector`), no Ingress, and internal-only status. The traffic-blocking check stays gated behind `OGX_E2E_TEST_NP_ENFORCEMENT=true` (only meaningful on a policy-enforcing CNI).

## Upgrade note

Breaking in Praxis mode on enforcing-CNI clusters: co-located workloads that reach OGX today (not matched by the Praxis selector, outside the operator namespace) lose access, and any external Ingress is removed. Escape hatches: set `spec.praxisMode.praxisSelector`, add additive `spec.network.policy.ingress` rules, disable the policy with `policy.enabled: false`, or run legacy mode (`spec.praxisMode.enabled: false`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This repo hosts a Kubernetes operator that creates and manages OGX (Open GenAI S
     - [Installation](#installation)
     - [Deploying the OGX Server](#deploying-the-ogx-server)
     - [Runtime Config via CR](#runtime-config-via-cr)
-- [Enabling Network Policies](#enabling-network-policies)
+- [Network Policies (internal-only / Praxis-fronted)](#network-policies-internal-only--praxis-fronted)
 - [Monitoring](#monitoring)
 - [Developer Guide](#developer-guide)
     - [Prerequisites](#prerequisites)
@@ -184,9 +184,57 @@ kubectl apply -f config/samples/example-with-configmap.yaml
 
 `spec.overrideConfig` always takes precedence over declarative generation fields.
 
-## Enabling Network Policies
+## Network Policies (internal-only / Praxis-fronted)
 
-Network policies are enabled by default per-CR. Configure via `spec.network.policy`:
+OGX is an **internal-only backend**: in the target topology it is fronted by Praxis, which is
+the sole public entrypoint. The operator can enforce this at the network layer via
+**Praxis-fronted mode**.
+
+### Praxis-fronted mode (`spec.praxisMode`)
+
+`spec.praxisMode.enabled` is a tri-state per-CR switch:
+
+- **`true`** — Praxis-fronted (internal-only): the locked-down NetworkPolicy below, and no
+  external exposure.
+- **`false`** — legacy behavior: the pre-Praxis NetworkPolicy peers (all pods in the same
+  namespace + the OpenShift router) and `network.externalAccess.enabled` is honored (an Ingress
+  is created when enabled).
+- **unset** — a brand-new (greenfield) install has `spec.praxisMode.enabled` defaulted to `true`
+  by the operator's **mutating admission webhook** on create, so it adopts Praxis-fronted mode. A
+  CR created before the webhook existed (i.e. across an operator upgrade) keeps an unset value,
+  which is treated as legacy so the upgrade does not silently cut off co-located workloads. Because
+  the default is applied per-CR at create time, different OGXServers can migrate independently.
+
+> The `spec.praxisMode` API (the `enabled`, `praxisSelector`, and `migrationJob` fields and the
+> mutating webhook defaulter) is introduced by [#355](https://github.com/ogx-ai/ogx-k8s-operator/pull/355)
+> (RHAIENG-7193); this operator wiring consumes it.
+
+The rest of this section describes **Praxis mode**. In Praxis mode, for every OGXServer the
+operator creates a `NetworkPolicy` whose ingress on the service port (`8321`) admits traffic
+**only** from:
+
+1. **Praxis pods** — identified per-CR by `spec.praxisMode.praxisSelector` (see below), defaulting
+   to pods labeled `app: payload-processing` (the MaaS Gateway contract). This is the only
+   *application* traffic OGX accepts.
+2. **The operator namespace** — so the operator can poll OGX status (`/v1/providers`,
+   `/v1/version`). This is control-plane traffic, not application traffic.
+
+The broad "all pods in the same namespace" rule and the OpenShift router rule are **not**
+included, and the operator does **not** create any external exposure (Ingress) for OGX.
+Setting `spec.network.externalAccess.enabled: true` is **not honored** in Praxis mode — it is
+treated as `false` and surfaced as an admission warning (rather than a hard rejection, so
+existing CRs and GitOps applies keep working). `status.serviceURL` is populated with the
+internal cluster DNS endpoint; `status.externalURL` is empty. (In legacy mode, external access
+is honored and `status.externalURL` reflects the created Ingress.)
+
+In Praxis mode the operator also disables the **Responses API** in OGX's generated config (it is
+served by Praxis instead). This is applied internally during config generation and does **not**
+mutate your CR's `spec.disabledAPIs`.
+
+### Adding your own ingress rules (additive)
+
+`spec.network.policy.ingress` rules are **appended on top of** the mandatory Praxis + operator
+rules — they cannot remove the lock-down:
 
 ```yaml
 apiVersion: ogx.io/v1beta1
@@ -197,9 +245,6 @@ spec:
   distribution:
     name: starter
   network:
-    externalAccess:
-      enabled: true
-      hostname: my-ogx.example.com
     policy:
       enabled: true
       ingress:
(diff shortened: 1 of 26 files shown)


eoinfennessy (commented):
The OGX NetworkPolicy should only be configured for "behind Praxis mode" when `praxis-mode: true` (see comment below for the why - TL;DR: it will break user systems on upgrade). When `praxis-mode: false` the NetworkPolicy behaviour should remain exactly the same as it is now. I think this work may be blocked until we have added the `praxis-mode` field (or decided on a praxis mode config struct - see conversations linked below).

Why are we configuring the Praxis peer through a ConfigMap? Should this be configurable through the OGXServer CR itself?

Is it possible that users could have multiple Praxis instances? I'm assuming this is the case. Having a way of specifying the correct Praxis instance to migrate to seems like something we might need.

Relevant conversations:
1. [On explicit migration using the `praxis-mode` field](https://docs.google.com/document/d/1W1ZwZSCoal3hOG-e0iGtUc8Tn6yx01ehunDQkLOrB7g/edit?disco=AAACFeh7yTE)
2. [On allowing users to specify the Praxis instance to migrate to](https://docs.google.com/document/d/1W1ZwZSCoal3hOG-e0iGtUc8Tn6yx01ehunDQkLOrB7g/edit?disco=AAACFxq8-8c)

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+  **all** namespaces.
+
+## Open questions
+
+### 1. Default direction on upgrade vs. product intent
+The product strategy (RHAISTRAT-2277, accepted one-pager) says "OGX should be internal-only."
+Eoin's review argued the switch must be opt-in so upgrades don't break enforcing-CNI clusters.
+We split the difference: **greenfield defaults ON, existing installs default OFF.** Is this the
+agreed behavior, or should Praxis mode be strictly opt-in (default OFF everywhere) until a later
+release flips the default?
```
eoinfennessy: > **greenfield defaults ON, existing installs default OFF.**

This is the correct behaviour. What is described in RHAISTRAT-2277 is not possible because unexpected downtime for users is inevitable.

How I think this could be implemented:
- have an optional pointer to a bool allowing us to distinguish between unset and explicitly set to true/false
- implement a mutating webhook that sets this to true for all new CRs (where event = create)

With this approach...
- all existing CRs, or new CRs that explicitly set the field to false will use the legacy OGX mode, and
- all new CRs, or existing CRs that explicitly set the field to true will use the legacy OGX mode

The current approach in this PR is not far off this, but it uses the CR's status to persist state that determines praxis mode resolution. Determining reconciliation behaviour from status is unconventional, and can cause other issues. I think using mutating webhooks to modify newly-created CRs is a more  conventional/predictable approach.

cc @VaishnaviHire

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 2. Is the greenfield-vs-upgrade heuristic acceptable?
+We detect "existing install" via `status.phase != ""` (a CR previously reconciled by any
+operator version) and persist the decision to `status.praxisModeResolved`. Edge cases to confirm:
+- A CR that was created but never successfully reconciled (phase still empty) is treated as
+  greenfield → ON. Acceptable?
+- Should the resolved mode instead be derived from an explicit install-time signal (e.g. a DSC
+  field or operator-wide default) owned by RHAIENG-6736, rather than inferred per-CR?
```
eoinfennessy: See comment above. We should not be using `status` of any kind to determine reconciliation behaviour.

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 3. Per-CR field vs. operator-wide vs. RHAIENG-6736
+The mode is a **per-CR field** so different OGXServers can migrate independently. RHAIENG-6736
+("OGX operator supports internal-only deployment mode") is the story that formally owns this
+concept and may prefer a different shape (operator-wide toggle, DSC-driven, different field name,
+router-peer tied to `externalAccess.enabled` per its AC#8). **Action:** reconcile this
+implementation with RHAIENG-6736's owner before merge so we don't ship two competing contracts.
```
eoinfennessy: > different OGXServers can migrate independently

I think this is the correct UX to offer. We should not expect users to migrate all OGXServers simultaneously - each instance could be managed by different teams, who may need to manage their migration independently.

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 4. Multiple Praxis instances per cluster
+`praxis-peer` is a single operator-wide selector. If a cluster can run **multiple Praxis
+instances** (e.g. per-tenant), a single cluster-wide peer cannot express "OGXServer A fronts
+Praxis X, OGXServer B fronts Praxis Y." Do we need a per-CR Praxis peer/reference? If so, the
+peer should probably move (or be overridable) on the CR alongside `praxisMode`.
```
eoinfennessy: This needs discussion with Praxis Core team. I asked about this [here](https://redhat-internal.slack.com/archives/C0B6RE27M0S/p1787650207720659?thread_ts=1787580398.100339&cid=C0B6RE27M0S).

Aside: the inverse of this also needs to be considered - if we are deploying a single praxis instance...
1. how will this proxy to multiple OGX instances?
2. how will data be managed for multiple OGX instance migrations?

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 5. Praxis namespace contract (selector's namespace half)
+The pod label `app: payload-processing` is agreed, but the **namespace** half of the selector is
+not. The fail-safe default has no `namespaceSelector`, so it admits that label in **any**
+namespace (documented as intentional, see README). Which namespace(s) does Praxis run in, and
+should we pin `namespaceSelector: {kubernetes.io/metadata.name: <ns>}` or a shared label by
+default?
```
eoinfennessy: The PoC operator puts extproc servers in the `openshift-ingress` namespace - needs follow up with MaaS folks (Ryan Qin, Jamie Land).

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 6. `disabledAPIs: ["responses"]` in greenfield Praxis mode
+RHAIENG-6736 notes that greenfield Praxis-fronted deployments should disable the `responses` API
+(served by Praxis instead). Should enabling Praxis mode also default `disabledAPIs` to include
+`responses`, or is that left to the CR author / a separate story?
```
eoinfennessy: Enabling Praxis mode should disable the Responses API. This should be applied internally by the controller either during or after the config resolution pipeline `pkg/config`. It should not mutate users' CRs.

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 7. Route / HTTPRoute coverage
+The operator only ever creates an `Ingress` for external access, so internal-only enforcement
+only removes Ingress. If OpenShift `Route` or Gateway API `HTTPRoute` exposure is added later,
+Praxis mode must suppress/remove those too. Confirm no external Route/HTTPRoute path exists today.
```
eoinfennessy: Removing user-managed networking should not be the responsibility of the operator.

This is another reason migration should be explicit - users may have additional networking pointing to OGX that they need to manually update to point at Praxis before migration.

docs/praxis-mode-open-questions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### 8. Explicit `praxisMode` override then removal
+If a user sets `spec.praxisMode` explicitly and later removes it, resolution falls back to
+`status.praxisModeResolved` (if previously persisted) or the heuristic. We do **not** currently
+persist a resolution while the spec value is explicit. Confirm this fallback behavior is what we
+want, or persist the effective mode on every reconcile.
```
eoinfennessy: See above comments re. not using status to determine reconciliation behaviour.

eoinfennessy (commented):
Added comments on the open questions.

I think we should block the NetworkPolicy stuff until the `praxis-mode` field (or whatever config options we decide on) has been added, and we determine whether or not we need to support multiple Praxis instances. WDYT?

docs/praxis-mode-open-questions.md
nathan-weinberg: Agreed

docs/praxis-mode-open-questions.md
nathan-weinberg: Ack - I will assume the `openshift-ingress` namespace for now, but we can update this later if needed (either to a new namespace selector or none if that is acceptable)

docs/praxis-mode-open-questions.md
nathan-weinberg: Agreed

PR #355, opened 2026-08-26
feat: add Praxis mode with mutating webhook defaulter

eoinfennessy:
## Summary

Adds `spec.praxisMode` to `OGXServer` CRD. This includes the API surface, validation logic, and defaulting logic only - it is not wired up to the controller in any way.

- New API types: `PraxisModeSpec` (`enabled`, `praxisSelector`, `migrationJob`), `PraxisSelector`, and `MigrationJobSpec`
- Mutating admission webhook + defaulter that enables Praxis mode on **CREATE only**, leaving pre-existing CRs untouched
- cert-manager and OpenShift overlays wire CA injection into the new `MutatingWebhookConfiguration`
- Regenerated CRD, webhook manifests, deepcopy, API docs, and release installers

## Behaviour

- Praxis mode is enabled by default for all newly-created OGXServer CRs; Existing CRs must explicitly enable Praxis mode.
- The DB migration Job is disabled for all newly-created OGXServer CRs; Existing CRs must explicitly enable the DB migration Job.
- Users have the option to specify a PostgreSQL instance to write migration data to. If unspecified, the existing OGX PostgreSQL instance will be used.
- Users have the option to select a specific Praxis instance to act as gateway. A default Praxis instance will be provided later - either through Go constants or a platform-level ConfigMap.

## Test plan

- `make test TEST_PKGS=./api/v1beta1` — passes, incl. new `TestCEL_PraxisSelector` (matchLabels-only, matchExpressions-only, empty-selector cases) and updated defaulter tests
- `make lint` — 0 issues
- `pre-commit run` — all hooks pass (manifests, install script, API docs regenerated and committed)


api/v1beta1/ogxserver_types.go
```diff
@@ -449,6 +449,56 @@ type WorkloadSpec struct {
...
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// TargetConnectionString references a Secret containing the PostgreSQL
+	// connection string that the migration Job writes to. When omitted, the
+	// Job writes to the same PostgreSQL instance it reads from (the connection
+	// string configured in spec.storage.sql.connectionString).
+	// The Secret must be in the same namespace as the OGXServer
+	// and must have the label ogx.io/watch: "true".
+	// +optional
+	TargetConnectionString *SecretKeyRef `json:"targetConnectionString,omitempty"`
```
VaishnaviHire, line 479: Do we want to expose `targetConnectionString` in the API?

The current migration discussion seems to be converging on an in-line migration within the configured OGX database. Exposing a separate target connection string makes cross-DB migration look like a supported path ?

 I think unless separate DB migration is an explicit requirement, we can remove this field and use the `spec.storage.sql.connectionString` as the migration DB connection for now.

api/v1beta1/ogxserver_types.go
```diff
@@ -449,6 +449,56 @@ type WorkloadSpec struct {
...
+
+	// PraxisSelector identifies the Praxis instance that is the gateway to the OGX server.
+	// Defaults to the cluster's default Praxis instance if omitted.
+	// +optional
+	PraxisSelector *PraxisSelector `json:"praxisSelector,omitempty"`
+
+	// MigrationJob configures the DB migration Job that migrates OGX data to Praxis.
+	// When omitted the Job is not created.
+	// +optional
+	MigrationJob *MigrationJobSpec `json:"migrationJob,omitempty"`
```
VaishnaviHire, line 499: For an existing CR, setting `spec.praxisMode.enabled=true` appears to opt into Praxis mode, but migration will not run unless `migrationJob` is also set. Enabling Praxis mode without data migration could be unsafe for brownfield deployments once controller wiring is added.  Maybe we mention this in docs or set this to `enabled` by default

VaishnaviHire (commented):
Thanks for the changes. The changes overall look good.

I am assuming the scope here is to only introduce the `praxis` mode updates and the controller changes for internal-only oxg deployment will be a follow-up.

api/v1beta1/ogxserver_types.go
eoinfennessy: I think there could be legitimate use cases for requiring a different PostgreSQL instance - e.g. organizational boundary requirements, or users that want to keep databases in the same namespace as the workloads that use them.

I think it's a cheap and easy change to make - i.e. swapping out one connection string secret for another of the same type if provided.

WDYT?

api/v1beta1/ogxserver_types.go
eoinfennessy: I kind of think data migration should probably be explicit too - any move of user data feels like something that should be opt in, especially given potential gaps in functionality when moving to Praxis (thinking of ABAC in particular). It's also possible users may want to enable praxis mode without migrating data - e.g. for a dry run, or to start afresh with a clean DB.

We will have to document the migration process. Adding something like the following in the migration docs would cover it:

> When enabling Praxis mode, if you wish to migrate your Responses and Conversations data from the OGX DB to the Praxis DB, you must also enable the data migration Job by setting `spec.praxisMode.migrationJob.enabled` to `true`.

api/v1beta1/ogxserver_types.go
VaishnaviHire: I think it should be okay to add this as an optional field. However, we should be explicit that the supported path is inline migration. 
cc: @leseb

api/v1beta1/ogxserver_types.go
nathan-weinberg: So IIUC the potential scenarios for brownfield deployments looking to upgrade here are:
1. praxis mode enabled, no data migration
2. praxis mode enabled, data migration job run
3. data migration job run only (though I assume this would not only be bad practice but potentially breaking and we might want to block?)

nathan-weinberg (approved):
this looks very good to me! chimed in on one convo - i think it would also be good to potentially add some more in-depth testing to https://github.com/ogx-ai/ogx-k8s-operator/tree/main/tests/e2e and/or https://github.com/opendatahub-io/opendatahub-tests at some point - we'll be building quite a bit off this somewhat quickly and want to reduce the risk of regressions as much as possible as we enter this brave new world

api/v1beta1/ogxserver_types.go
eoinfennessy: Yes for 1 and 2; no for 3 - the migration job will not run if Praxis mode is disabled.

eoinfennessy: ...we should probably add validation to make 3 unrepresentable. This may be possible with CEL - failing that, we can use validating webhooks

nathan-weinberg: Yes definitely!

PR #360, opened 2026-08-31
feat: add Praxis-fronted internal-only deployment mode for OGXServer

nathan-weinberg:
## What

Adds an explicit **Praxis-fronted, internal-only deployment mode** to the OGX operator. OGX is an internal backend fronted by Praxis (the public entrypoint); in this mode the operator restricts OGX so it cannot be reached directly (bypassing Praxis auth/tenant normalization) or exposed externally.

This PR delivers **RHAIENG-6736** (internal-only deployment mode), which now subsumes **RHAIENG-6663** (Praxis-only NetworkPolicy ingress).

> **Targets the `feature/praxis` branch.** The `spec.praxisMode` API (`enabled`, `praxisSelector`, `migrationJob`) and the mutating webhook defaulter (RHAIENG-7193, merged as #359) already live on `feature/praxis`; this PR is stacked on top and only wires the operator to consume it.

## Mode selection (`spec.praxisMode.enabled`, upgrade-safe)

Tri-state, resolved per-CR — no reconciliation state is derived from `status`:

- **`true`** — Praxis-fronted (internal-only).
- **`false`** — legacy behavior.
- **unset** — new CRs get `enabled` defaulted to `true` on create by the mutating webhook (greenfield → Praxis mode); CRs predating the webhook keep an unset value and are treated as **legacy**, so an operator upgrade never silently cuts off co-located workloads. Because defaulting is per-CR at create time, OGXServers migrate independently.

## Changes

- **Consume the `spec.praxisMode` API** — dropped this branch's own `spec.praxisMode *bool` field, its mutating defaulter, and the operator-wide `praxis-peer` ConfigMap mechanism in favor of the `spec.praxisMode` struct.
- **Per-CR Praxis peer** — the NetworkPolicy Praxis ingress peer is built from **`spec.praxisMode.praxisSelector`** (namespace + pod selector), so different OGXServers can front different Praxis instances (addresses @VaishnaviHire and @eoinfennessy's per-CR / multiple-Praxis-instances point). Falls back to `app: payload-processing` in `openshift-ingress` when unset; never emits an allow-all peer (the selector is CEL-validated non-empty).
- **Praxis mode NetworkPolicy** — ingress on `8321` admits only the Praxis peer + the operator namespace (control-plane status polling). The broad same-namespace peer and OpenShift-router peer are removed. `spec.network.policy.ingress` rules are **additive** on top of the mandatory lock-down. `policy.enabled: false` still fully disables the policy.
- **Enforced internal-only external access (Praxis mode)** — the operator never creates an Ingress and removes any it previously created; `status.externalURL` is cleared while `status.serviceURL` keeps the internal cluster DNS endpoint. `spec.network.externalAccess.enabled: true` is not honored and surfaces an admission **warning** (not a rejection).
- **Responses API** — disabled in the generated config in Praxis mode (served by Praxis); applied internally, does not mutate `spec.disabledAPIs`.
- **Legacy mode** — preserves the pre-Praxis peers (same-namespace + OpenShift router) and `externalAccess`-driven Ingress (RHAIENG-6736 AC#8: router peers apply only in legacy mode).

Out of scope here (tracked by RHAIENG-7193): wiring the DB **migration Job** (`spec.praxisMode.migrationJob`) and the target-PostgreSQL instance.

## Testing

- Unit / envtest (all non-e2e packages pass): `resolvePraxisMode` and `BuildPraxisPeer` (praxisSelector→peer mapping + default fallback, never allow-all); NetworkPolicy contains only Praxis + operator peers in Praxis mode and same-namespace + router in legacy; user rules merge additively; no Ingress created even with `externalAccess.enabled: true`; previously-owned Ingress deleted; `status.serviceURL` populated / `status.externalURL` empty; webhook warns (not errors) on `externalAccess.enabled: true`; Responses API disabled in generated config.
- CI-runnable e2e negative suite asserts on-cluster NetworkPolicy shape (now driven by per-CR `praxisSelector`), no Ingress, and internal-only status. The traffic-blocking check stays gated behind `OGX_E2E_TEST_NP_ENFORCEMENT=true` (only meaningful on a policy-enforcing CNI).

## Upgrade note

Breaking in Praxis mode on enforcing-CNI clusters: co-located workloads that reach OGX today (not matched by the Praxis selector, outside the operator namespace) lose access, and any external Ingress is removed. Escape hatches: set `spec.praxisMode.praxisSelector`, add additive `spec.network.policy.ingress` rules, disable the policy with `policy.enabled: false`, or run legacy mode (`spec.praxisMode.enabled: false`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,9 +227,24 @@ existing CRs and GitOps applies keep working). `status.serviceURL` is populated 
 internal cluster DNS endpoint; `status.externalURL` is empty. (In legacy mode, external access
 is honored and `status.externalURL` reflects the created Ingress.)
 
-In Praxis mode the operator also disables the **Responses API** in OGX's generated config (it is
-served by Praxis instead). This is applied internally during config generation and does **not**
-mutate your CR's `spec.disabledAPIs`.
+In Praxis mode the operator also disables the **Responses API** and **Conversations API** in OGX's
+generated config (they are served by Praxis instead). This is applied internally during config
+generation and does **not** mutate your CR's `spec.disabledAPIs`.
+
+#### Praxis-mode readiness conditions
+
+Because OGX and Praxis are deployed independently, the operator surfaces two Praxis-mode
+preconditions as status conditions rather than admission rejections (the operator does not manage
+Praxis, and a Praxis instance or TLS Secret may legitimately appear after OGX):
+
+- **`PraxisReachable`** — `True` when the effective Praxis selector (`spec.praxisMode.praxisSelector`,
+  or the fail-safe default) resolves to at least one Ready Praxis pod. When `False`, internal-only
+  OGX has no valid path until Praxis is available.
+- **`TLSConfigured`** — `True` when `spec.network.tls.secretName` is set and the referenced Secret
+  exists (it must carry the `ogx.io/watch: "true"` label to be detected). When `False`, OGX is not
+  serving its internal endpoint over the expected mTLS.
+
+These conditions are only evaluated in Praxis mode.
 
 ### Adding your own ingress rules (additive)
 
@@ -258,7 +273,10 @@ spec:
 ```
 
 To fully disable the NetworkPolicy (emergency escape hatch), set
-`spec.network.policy.enabled: false`.
+`spec.network.policy.enabled: false`. In Praxis mode this removes the mandatory Praxis + operator
+lock-down entirely (OGX may then be reachable directly by co-located workloads, bypassing Praxis),
+so it is surfaced as an admission warning. Prefer additive `spec.network.policy.ingress` rules
+instead.
 
 ### Configuring the Praxis peer (per-CR)
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This repo hosts a Kubernetes operator that creates and manages OGX (Open GenAI S
     - [Installation](#installation)
     - [Deploying the OGX Server](#deploying-the-ogx-server)
     - [Runtime Config via CR](#runtime-config-via-cr)
-- [Enabling Network Policies](#enabling-network-policies)
+- [Network Policies (internal-only / Praxis-fronted)](#network-policies-internal-only--praxis-fronted)
 - [Monitoring](#monitoring)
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -212,12 +257,46 @@ spec:
               port: 8321
 ```
 
+To fully disable the NetworkPolicy (emergency escape hatch), set
+`spec.network.policy.enabled: false`.
+
```
VaishnaviHire, line 280: Since `spec.network.policy.enabled=false` fully disables the Praxis-mode NetworkPolicy, should we surface an admission warning similar to the externalAccess warning?

README.md
```diff
@@ -184,9 +184,57 @@ kubectl apply -f config/samples/example-with-configmap.yaml
...
+The broad "all pods in the same namespace" rule and the OpenShift router rule are **not**
+included, and the operator does **not** create any external exposure (Ingress) for OGX.
+Setting `spec.network.externalAccess.enabled: true` is **not honored** in Praxis mode — it is
+treated as `false` and surfaced as an admission warning (rather than a hard rejection, so
+existing CRs and GitOps applies keep working). `status.serviceURL` is populated with the
+internal cluster DNS endpoint; `status.externalURL` is empty. (In legacy mode, external access
+is honored and `status.externalURL` reflects the created Ingress.)
+
+In Praxis mode the operator also disables the **Responses API** in OGX's generated config (it is
```
VaishnaviHire: We should also disable `Conversations` API. This will require changes in operator spec as well as upstream ogx server. If you want to handle this as a separate follow-up, that works as well.

VaishnaviHire (commented):
Thanks for the updates @nathan-weinberg .

README.md
nathan-weinberg: I think so yes, good idea!

README.md
nathan-weinberg: I will go ahead and include it in this PR, good call!

VaishnaviHire (approved):

rhdedgar (approved):
Looks like this would cover RHAIENG-6644 as well, +1 to merge.

PR #367, opened 2026-09-09
feat: OGX to Praxis upgrade migration job and cutover gates

VaishnaviHire:
## Summary

When `spec.praxisMode.enabled` and `spec.praxisMode.migrationJob` are set, the operator runs preflight, creates Job `{name}-praxis-migration` (`ogx migrate praxis --dry-run` then `ogx migrate praxis`,
[opendatahub-io/ogx#233](https://github.com/opendatahub-io/ogx/pull/233)), and sets `PraxisCutoverReady` only after Job `Complete`. CRs that never opt in get no Job and no `status.migration`.

## Changes

- Preflight requires the OGX Postgres Secret (`spec.storage.sql.type=postgres`
  and `connectionString` with `ogx.io/watch=true`).
- `targetConnectionString` is only for a separate Praxis database; omit it to
  write to the same OGX Secret.
- Config: mount the operator-generated ConfigMap at `/etc/ogx/config.yaml` when
  the server would mount one; otherwise use the image file at that path. A user
  ConfigMap is not required.
- Credentials via `SecretKeyRef` only. Job uses the server SA / FSGroup / resources.
- Attempt key is Secret content + mounted ConfigMap content (when present) + image.
  Same attempt is not duplicated.
- Terminal state is Job conditions `Complete` / `Failed`, not `status.failed > 0`.
  Pod retries only (`backoffLimit=2`). A Failed Job is kept for `oc logs`.
- Retry: delete the Failed Job (or change Secret/config/image). Replace waits
  until the old Job is gone. Do not delete a succeeded Job — Validated is not
  recreated.
- Soft rollback only: durable warning + one Event on disable. No DB rollback.
- Docs: `docs/upgrade-praxis-migration.md`.

## Test plan

- [ ] `make lint` / `make test`
- [ ] OpenShift: opt-in brownfield CR with OGX Postgres Secret →
      `PraxisCutoverReady=True`; fail → Job remains and `oc logs` works; delete
      Job to retry; disable after a run → one soft-rollback warning. Image must
      contain `ogx migrate praxis`. Freeze writers operationally (no API freeze).


controllers/migration_job.go
```diff
@@ -0,0 +1,722 @@
...
+	return nil
+}
+
+func (r *OGXServerReconciler) createMigrationJob(
+	ctx context.Context,
+	instance *ogxiov1beta1.OGXServer,
+	runtimeConfig *runtimeConfigRef,
+	jobName, attemptKey string,
+) (*batchv1.Job, error) {
+	job, err := r.buildMigrationJob(instance, runtimeConfig, jobName, attemptKey)
```
rhdedgar: We might want to add some TLS check for configured CA bundles and add the same CA bundle volume, volume mount, and SSL_CERT_FILE env var that the server Deployment gets.

controllers/migration_job.go
```diff
@@ -0,0 +1,722 @@
...
+		return false
+	}
+	mj := instance.Spec.PraxisMode.MigrationJob
+	if mj == nil {
+		return false
+	}
+	return mj.Enabled == nil || *mj.Enabled
+}
+
+func (r *OGXServerReconciler) handleMigrationNotRequested(instance *ogxiov1beta1.OGXServer) {
```
rhdedgar: We might also want to look up and delete the existing migration `Job` via `DeletePropagationBackground` before clearing status. So that it stops the `Job` explicitly.

rhdedgar (approved):
+1, left a couple comments but they're not something I would block the PR over.

api/v1beta1/ogxserver_types.go
derekhiggins: I think we need to make this compulsory, it can't default to the same PostgreSQL instance it reads from, the default OGX source and Praxis target both contain  table named `openai_conversations`

eoinfennessy: Yes, that's a good point. We should either make it compulsory when migration is enabled, or if not provided change the database name part of the connection string to be `praxis` (or whatever the default DB name Praxis currently uses). I'm leaning towards the former to keep things simple / explicit.

controllers/migration_job.go
VaishnaviHire: Addressed: the Job now mounts the same managed CA bundle volume as the server and sets `SSL_CERT_FILE` when TLS trust (or an ODH trusted CA) is configured. Preflight also requires that ConfigMap to exist before the Job is created.

api/v1beta1/ogxserver_types.go
VaishnaviHire: Went with making it compulsory when migration is enabled, as you suggested.

controllers/migration_job.go
VaishnaviHire: Addressed: when migration is not opted in, the operator deletes the existing `{name}-praxis-migration` Job with `DeletePropagationBackground` before clearing status.

api/v1beta1/ogxserver_types.go
VaishnaviHire: Agreed — `targetConnectionString` is now required when `migrationJob` is set (CRD + CEL + preflight). There is no fallback to the OGX source database.

VaishnaviHire: Good catch, fixed it.

api/v1beta1/ogxserver_types.go
```diff
@@ -474,25 +474,31 @@ type PraxisSelector struct {
...
-	TargetConnectionString *SecretKeyRef `json:"targetConnectionString,omitempty"`
+	// +kubebuilder:validation:Required
+	TargetConnectionString *SecretKeyRef `json:"targetConnectionString"`
 }
 
 // PraxisMode configures integration with an existing Praxis instance that
 // acts as gateway for this OGX server.
+// +kubebuilder:validation:XValidation:rule="!has(self.migrationJob) || has(self.migrationJob.targetConnectionString)",message="targetConnectionString is required when migrationJob is set"
+//
+//nolint:lll // kubebuilder markers cannot be split across lines.
```
eoinfennessy: This CEL validation can be removed because `TargetConnectionString` is now a required field.
```suggestion
```

controllers/kubebuilder_rbac.go
```diff
@@ -46,6 +46,13 @@ package controllers
 // PodDisruptionBudget permissions - controller creates and manages voluntary disruption controls
 //+kubebuilder:rbac:groups=policy,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
 
+// Job permissions - controller creates and tracks Praxis migration Jobs
+//+kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=batch,resources=jobs/status,verbs=get
+
+// Event permissions - controller emits migration progress/failure events
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
```
eoinfennessy, line 54: Why do we need to emit events?

If we want to warn users of soft rollback (i.e. no Praxis DB writes migrated to OGX DB) we could instead add a validating webhook warning (not a hard error) when `spec.praxisMode.enabled` is updated from `true` to `false`.

eoinfennessy (commented):
Generally looks good to me. Two suggestions inline and a couple of agent findings below:

High: replacement migrations may overlap. Stale Jobs are deleted with background propagation, so the Job can disappear before its Pods terminate; reconciliation may then start the replacement while the old migration still runs. Use foreground deletion or verify dependent Pods are gone. See [migration_job.go (line 498)](/Users/efenness/github/ogx-ai/ogx-k8s-operator/controllers/migration_job.go:498).

Medium: source and target may be the same database. Preflight verifies both Secrets exist but never rejects identical references or connection strings, despite the API warning that they must differ. See [migration_job.go (line 184)](/Users/efenness/github/ogx-ai/ogx-k8s-operator/controllers/migration_job.go:184).

Medium: an unowned Job can be trusted or deleted. Existing Jobs are checked only by name and attempt annotation, without verifying their controller owner UID. A leftover Job from a recreated CR could be treated as successful—or deleted as stale. See [migration_job.go (line 451)](/Users/efenness/github/ogx-ai/ogx-k8s-operator/controllers/migration_job.go:451).

controllers/kubebuilder_rbac.go
VaishnaviHire: Yeah I will move the rollback warning to webhook. But I think the events will still be helpful for Job create / fail / cutover. Those happen after admission, and they show up on `oc describe` without scraping conditions.

VaishnaviHire: Done

derekhiggins (approved):
lgtm, thanks

nathan-weinberg (commented):
Reviewed at `d7fddf5`. Branch checked out locally: `make lint` clean, `make manifests generate api-docs` produces no diff, and the 24 new migration tests pass. `TestReconcile` fails, but it fails identically on `feature/praxis` (`failed to resolve base config from OCI labels: OCI label fetcher is not configured`) — pre-existing, not caused by this PR.

Findings 1, 2, 5, and 6 were confirmed empirically with throwaway tests written against this PR's own fixtures (`migrationInstance`, `migrationObjects`, `runtimeCfg`, `markJobComplete`), then deleted. Raw output is inlined under each finding.

## Overall

Solid, careful work. Things done well:

- Opt-in design with **zero status footprint** for CRs that never participate.
- Terminal state read from Job **conditions** (`Complete`/`Failed`) rather than `status.failed > 0`.
- Foreground propagation on replace, so a new attempt waits for the old Pods to clear.
- `docs/upgrade-praxis-migration.md` is genuinely good — more accurate than the PR description.
- The `SetCondition` `LastTransitionTime`-preservation fix is a good general improvement worth having independent of this feature.
- Broad test coverage (24 cases), lint clean, generated manifests in sync.

The concerns are concentrated in one place:

> **The attempt key is used both as a retry-dedup mechanism and as a success invalidator, and the second role is unsafe for a one-shot data migration.**

## Concerns

### 1. A *successful* migration re-runs when config / secret / image content changes

`ensureMigrationJob` → `reconcileExistingMigrationJob` deletes any Job whose `ogx.io/migration-attempt` annotation doesn't match the current key (`controllers/migration_job.go:508`), with **no check for whether that Job already reached `Complete`**. `migrationAttemptAlreadySucceeded` is only consulted when the Job is *absent* (`controllers/migration_job.go:490`).

The `cfg=` component of the attempt key is the operator-generated ConfigMap, whose **name embeds a content hash**:

```go
// controllers/ogxserver_controller.go:676
pendingGeneratedConfigMapName = generatedConfigMapName(instance.Name, generated.ContentHash)
```

So any post-cutover spec change that alters the rendered `config.yaml` — adding an inference model, tweaking a provider — invalidates the key.

Confirmed:

```
after success:                  phase=Validated attempt=1a1d4f31282c2aa7
after config change (pass1):    phase=Running  cutoverReady=false
succeeded job still present?    jobs.batch "demo-praxis-migration" not found
NEW job attempt=d38056ee1a580c4e
```

**Impact:** a routine model addition weeks after cutover deletes the audit-trail Job, flaps `PraxisCutoverReady` to `False`, and re-runs `ogx migrate praxis` against a live Praxis DB. The same applies to a source-DB password rotation (`src=` hashes secret *content*) and to an operator upgrade that bumps the distribution image.

Docs line 66 says *"Changing Secret/config/image content also starts a new attempt"* — so this is intentional, but the intent doesn't survive contact with `Validated`. That rule is right for a *pending or failed* attempt and wrong for a succeeded one.

**Suggested fix:** once `status.migration.phase == Validated`, treat it as terminal. Only allow a new attempt when the Job is absent or `Failed`, or behind an explicit force annotation.

### 2. Disable → re-enable re-runs the migration with an identical key

`clearMigrationState` (`controllers/migration_job.go:138`) replaces the whole `MigrationStatus`, wiping `AttemptKey` and `JobName`, and `handleMigrationNotRequested` deletes the Job unconditionally — including a succeeded one (this PR's own `TestReconcileMigration_SoftRollbackWarning` asserts the deletion).

Flipping `migrationJob.enabled: false` → `true` with nothing else changed:

```
after success:    phase=Validated attempt=1a1d4f31282c2aa7
after disable:    phase=Pending attemptKey=""
after re-enable:  NEW job created, attempt=1a1d4f31282c2aa7 (same key as succeeded run: true)
```

Identical inputs, identical key, and it still re-ran — because both the memory of success and the Job that proved it were destroyed.

This directly contradicts `docs/upgrade-praxis-migration.md:71`:

> *"Do not delete a succeeded Job; that attempt stays Validated and is not recreated."*

**Suggested fix:** preserve `AttemptKey`/`Phase` across `clearMigrationState`, and skip deletion for a Job that reached `Complete`.

## Medium

### 3. The unowned-Job guard doesn't fire in production

`main.go:132` filters cached Jobs by `app.kubernetes.io/managed-by=ogx-operator`. A foreign Job named `<cr>-praxis-migration` won't carry that label, so:

1. `getMigrationJob` returns `nil, nil` (cache filtered it out)
2. `assertOwnedMigrationJob` is never reached
3. `Create` → `AlreadyExists` → re-`Get` (still filtered) → `nil`
4. → error at `controllers/migration_job.go:580`: *"already exists but could not be read"*

Docs promise the Job is "left untouched"; reality is a confusing hard failure.

`TestReconcileMigration_UnownedJobIsRejected` passes only because `fake.NewClientBuilder()` applies no label filter — the fixture Job at `controllers/migration_job_test.go:460` has no labels at all.

**Suggested fix:** use `r.APIReader` for the existence/ownership check, and add the managed-by label to the test fixture so it exercises the guarded path.

### 4. Migration errors escalate to whole-CR failure

`reconcileMigration` returns hard errors for the unowned-Job case and for any Secret/ConfigMap read failure. That propagates through `reconcileResources` into `updateStatus`, which sets `Phase: Failed` and `DeploymentReady=False` (`controllers/ogxserver_controller.go:1209`) — on a perfectly healthy server — plus exponential-backoff requeue.

The preflight path already models the better behavior: set conditions, return `nil`. Worth extending that to these cases.

### 5. `RestartPolicy: OnFailure` works against the "keep the Job for `oc logs`" design

```
restartPolicy=OnFailure backoffLimit=2
```

With `OnFailure` the **same Pod restarts in place**, so `oc logs job/<name>` shows only the final attempt; earlier ones need `--previous` and the first is gone entirely after two restarts.

`RestartPolicy: Never` creates one Pod per attempt and retains all three log sets — which is what the retry story in the docs actually wants.

### 6. Preflight can't validate the no-ConfigMap path

When `runtimeConfig == nil`, the Job still gets `OGX_CONFIG=/etc/ogx/config.yaml` with nothing mounted there:

```
OGX_CONFIG="/etc/ogx/config.yaml" volumes=0 volumeMounts=0
```

Preflight passes and the failure surfaces as an opaque CLI error at runtime.

**Suggested fix:** either drop `OGX_CONFIG` in that case and let the CLI use its own default, or document that the image must ship that file.

## Minor

- **PR description is stale in two places** (the docs file is correct — just sync the description):
  - *"targetConnectionString is only for a separate Praxis database; omit it to write to the same OGX Secret"* — the field is now `+kubebuilder:validation:Required` (`api/v1beta1/ogxserver_types.go:493`) and `ensureDistinctMigrationDatabases` rejects the same DB.
  - *"durable warning + one Event on disable"* — no Event is emitted on disable. The soft-rollback signal is an **admission warning only**; `TestReconcileMigration_SoftRollbackWarning` explicitly asserts that no such Event fires.
- `targetConnectionString` optional → required is a **breaking CRD change**. Fine on an unreleased feature branch, but worth a line in the description since it rejects existing `migrationJob: {}` CRs.
- `MigrationPhaseSucceeded` is declared, commented as unused, and still in the CRD enum. Drop it or use it.
- `MigrationValidated` and `MigrationJobSucceeded` are both set from the same `JobComplete` signal and can never disagree. If `MigrationValidated` is a placeholder for a future real validation step, saying so in the comment would help; otherwise consider collapsing them.
- `warnSoftRollbackOnPraxisDisable` (`controllers/migration_job.go:841`) re-checks `isMigrationRequested`, which its only caller already guarantees is `false`. Dead branch.
- `emitMigrationEvent` passes `reason` as the `action` argument to `Eventf` (`controllers/migration_job.go:838`); the `k8s.io/client-go/tools/events` API expects a verb there (`"Creating"`, `"Migrating"`).
- `applyMigrationPodDefaults` → `configurePodOverrides` appends `workload.overrides.volumes` to the migration pod. If a user override references an RWO PVC already bound to the server Deployment, the Job deadlocks on multi-attach — the same hazard the `Recreate` strategy exists to avoid elsewhere in this operator.

## Bottom line

**#1 and #2 are worth resolving before merge** — they turn a one-shot data migration into something that can silently re-fire during steady-state operation. Everything else is refinement.

One open question: `PraxisCutoverReady` looks *informational only* — nothing in the reconciler holds back the Deployment until the migration completes (`reconcileMigration` is the last step in `reconcileResources`, after the Deployment is applied). The title says "cutover gates"; can you confirm the gate is advisory by design?

VaishnaviHire:
Addressed `#1` , `#2` of the comments

`#4` (migration errors escalate to whole-CR failure): For this, I will open a follow-up, since this should updated across other migration paths as well.

`#5` Restart policy- kept it `onFailure` , we can document to  use `--previous` to get earlier logs.

nathan-weinberg (approved):

eoinfennessy (approved):

PR #368, opened 2026-09-10
feat: configure Praxis auth and tenancy

derekhiggins:
Generate runtime configuration for Praxis instances with or without declarative config. Replace the base auth provider with upstream-header authentication, apply the ownership access policy, and enable multi-tenant mode while preserving unknown default config sections.

Honor network port overrides in the default-config path and align configgen with the controller behavior.


rhdedgar (approved):
+1 here, it covers the operator requirements from RHAIENG-6642.

PR #373, opened 2026-09-16
fix(webhook): remove Praxis defaulting webhook

eoinfennessy:
## Summary

- remove the mutating defaulting webhook that enabled Praxis mode for all new OGXServer CRs
- add warning to validating webhook clarifying support level when Praxis mode is enabled:
	> Praxis mode is enabled (spec.praxisMode.enabled: true). The Responses API served by Praxis is Tech Preview.


api/v1beta1/ogxserver_types.go
```diff
@@ -495,7 +495,7 @@ type MigrationJobSpec struct {
 // acts as gateway for this OGX server.
 type PraxisModeSpec struct {
 	// Enabled controls whether Praxis mode is enabled.
-	// Defaults to true.
+	// Defaults to true when PraxisMode is provided.
 	// +optional
 	// +kubebuilder:default:=true
```
VaishnaviHire, line 500: Shouldn't this be `false`? We want users to explicitly enable this.

eoinfennessy: Praxis mode is disabled by default. If `spec.praxisMode` is provided it is then enabled by default, with an option for users to explicitly disable it. This follows the semantics of other fields in the spec - i.e. disabled if not provided; enabled by default if provided

nathan-weinberg (commented):
## Verification done

- `go build ./...`, `make lint` — clean.
- `make build-installer` — no drift; `config/webhook/manifests.yaml`, both `release/*.yaml`, CRDs and deepcopy are all regenerated consistently.
- Grepped the whole tree: no dangling `MutatingWebhookConfiguration` / `mogxserver.kb.io` / `/mutate-` references. The kustomize overlay surgery (cert-manager patch + both replacements, openshift patch, `kustomizeconfig.yaml`) is complete.
- Confirmed via envtest that the CRD structural default is doing the job the webhook used to: `praxisMode: {}` → `enabled: true` on create; `praxisMode` omitted → stays `nil` (legacy). So the new documented semantics are accurate.
- `make test TEST_PKGS="./api/v1beta1 ./controllers"`: `api/v1beta1` passes. `controllers` has one failure, `TestReconcile` ("OCI label fetcher is not configured") — it fails identically on `feature/praxis`, so it's pre-existing and unrelated to this PR.

## Findings

### 1. `cmd/configgen` now disagrees with the cluster for the new idiomatic spelling — `cmd/configgen/main.go:112`

`isPraxisMode` requires `Enabled != nil`, but configgen loads the CR from a YAML file, so no CRD schema defaulting ever runs. Now that `praxisMode:` provided-without-`enabled` is the documented way to turn Praxis mode on, the offline tool and the operator disagree on the same CR. Reproduced:

```yaml
spec:
  distribution: {name: starter}
  praxisMode:
    praxisSelector: {namespace: openshift-ingress, podSelector: {matchLabels: {app: payload-processing}}}
```

```
$ go run ./cmd/configgen -base base.yaml cr.yaml
failed to generate config: CR has no declarative config fields ...; nothing to generate
```

Adding `enabled: true` produces the full Praxis config. In-cluster, the first CR *is* Praxis mode. Fix is to mirror the CRD default:

```go
return server.Spec.PraxisMode != nil &&
    (server.Spec.PraxisMode.Enabled == nil || *server.Spec.PraxisMode.Enabled)
```

The comment above it is also stale — "this does not apply webhook defaults" should now read "CRD schema defaults". The same stale-semantics note applies to `controllers/network_resources.go:44`: the bullet "`spec.praxisMode.enabled` unset → legacy behavior" is now unreachable for anything coming from the API server, and contradicts the CRD default. Harmless in the controller (objects are always defaulted), but worth aligning the two helpers so they can't drift.

### 2. Nothing guards the CRD default that replaced the webhook

`TestOGXServerDefaulter_Default` and `TestMutatingWebhookOnlyFiresOnCreate` are gone, and no test took over. The entire "greenfield gets Praxis mode when `praxisMode` is provided" behavior now rests on a single `+kubebuilder:default:=true` marker, and dropping that marker would silently flip behavior with a green suite. `ogxserver_cel_test.go` already runs against envtest — two cases there (`praxisMode: {}` → `enabled == true`, `praxisMode` omitted → `PraxisMode == nil`) would restore the guard cheaply.

### 3. Stale comments the cleanup missed

- `tests/e2e/network_policy_test.go:48` — "the deployed webhook would default this on create, but be explicit so the suite is deterministic regardless of webhook availability". No longer true; the CR must set it explicitly.
- `controllers/ogxserver_controller_test.go:351` — `WithPraxisMode(true), // greenfield posture; webhook does not run in envtest`. The sibling comment in `testing_support_test.go` was updated; this one wasn't.

### 4. Nit — vacuous first case in `TestValidate_ExternalAccessWarning`

`"no network spec: no external access warning"` leaves `praxisMode` nil, so `collectValidationWarnings` returns early and the case can't distinguish "no network spec" from "not in Praxis mode". The analogous case in `TestValidate_PolicyDisabledWarning` correctly sets `praxisMode` enabled — worth matching.

### 5. Nit — README prose drops the omitted case

The bullet list under "Praxis-fronted mode" now only documents `true` and `false`; the "omit `praxisMode` for legacy" statement survives only in the field table below. A third bullet would keep the list self-contained.

## On the behavior change itself

This reverses #355's secure-by-default posture — new installs are legacy (co-located pods reachable, `externalAccess` honored) unless they opt in. That's clearly the PR's stated intent and the docs were updated honestly everywhere, so flagging it only so it's a conscious call: anything downstream that relied on greenfield CRs coming up internal-only now has to set `spec.praxisMode` explicitly.

Nothing here blocks the removal itself, which is clean. Item 1 is the one I'd fix before merge.

eoinfennessy:
@nathan-weinberg, Thanks for the review. I addressed points 1 to 4 in commits [b7ae65c](https://github.com/ogx-ai/ogx-k8s-operator/pull/373/commits/b7ae65cf2862322c61530c4a2466f195ad501d66) through [9ff4462](https://github.com/ogx-ai/ogx-k8s-operator/pull/373/commits/9ff4462548e9568afd2fefa0fb25b3a64feb16eb) respectively

nathan-weinberg (approved):

VaishnaviHire (approved):

PR #374, opened 2026-09-17
test: Praxis mode is opt-in and OGX Responses stays reachable by default (RHAIENG-6602)

nathan-weinberg:
Tests for [RHAIENG-6602](https://redhat.atlassian.net/browse/RHAIENG-6602).

## What changed, and why this PR looks different than it did

This PR originally proved the opposite of what it proves now.

Its first version was built on the premise that a greenfield install is Praxis-fronted and therefore must *not* expose OGX Responses. [RHAIENG-7517](https://redhat.atlassian.net/browse/RHAIENG-7517) / #373 removed that premise: Responses in Praxis will not be fully ready in 3.6, so OGX has to keep serving Responses standalone, and every OGXServer CR — new and existing — now defaults to Praxis mode **disabled**. A greenfield deployment is expected to expose OGX Responses, not hide it.

So the scope inverted, but most of the work didn't: the old assertions were never wrong, they just described the opt-in topology rather than the default one. They're reframed here, and the default topology gets its own coverage on both sides of every predicate.

Rebased onto `feature/praxis`, which already includes #373. Supersedes #372, which closed automatically when its base branch (`fix/remove-praxis-defaulting-webhook`) was deleted after #373 merged.

## Coverage

The five items in the ticket, and where each lands:

| Scope item | Where |
|---|---|
| A new CR defaults to Praxis mode disabled | `api/v1beta1/praxis_default_test.go` (static), `controllers/praxis_default_test.go` (envtest), `tests/e2e/greenfield_default_test.go` (live webhook) |
| An existing CR upgraded to 3.6 retains Praxis mode disabled | `TestUpgradedCR_KeepsPraxisModeUnsetAcrossUpdates`, plus the live `across an update` subtest |
| OGX Responses is reachable in the default topology | `TestDefaultCR_GeneratedConfigKeepsResponses` + the live `OGX serves /v1/responses` probe and the Ingress/NetworkPolicy assertions |
| The Tech Preview warning is emitted on an explicit opt-in | `tests/e2e/greenfield_default_test.go`, against the deployed webhook (unit-level coverage already landed in #373) |
| Routing/exposure when a user opts in, response shape stays OpenAI-compatible | `tests/e2e/praxis_optin_test.go` + `controllers/external_exposure_test.go` |

The "done when" clauses map to concrete failures: the default flipping back to enabled fails the static CRD/webhook guards and the live admission checks; Responses disappearing from the default topology fails both the generated-config assertion and the live probe; dropping the Tech Preview warning fails the e2e warning subtest.

### Three layers, deliberately

- **Static** (`api/v1beta1`) — reads the generated CRD, the controller-gen webhook output, and *both* `release/` installers. Milliseconds, cannot flake, fails on a `make manifests` diff before any cluster exists. A stale installer would keep registering a mutating webhook in a real cluster however clean `config/` looks, which is why the installers are read too.
- **envtest** (`controllers`) — drives the real reconcile loop against a real apiserver. No admission webhooks run there, which is exactly why it's the right place to isolate the upgrade case: anything that sets `spec.praxisMode` in envtest was set by the operator, not by admission.
- **e2e** (`tests/e2e`) — the deployed operator with its webhooks actually registered, and a live OGX pod answering HTTP.

## Not-vacuous by construction

Most of the review cost in a suite like this is deciding whether a passing test could ever fail. Each assertion is paired:

- `controllers/praxis_default_test.go` and `controllers/external_exposure_test.go` assert the **negation** of each other, through the same helpers. The router peer and the Ingress one file forbids are the ones the other requires. Neither can pass on a reconciler that stopped distinguishing the modes.
- The e2e default and opt-in suites share one external-path detector. The default suite requires it find a path; the opt-in suite requires it find none. A detector that silently stopped firing fails the former before it can quietly pass the latter.
- Both webhook guards check for the *validating* webhook before concluding anything from the absence of a *mutating* one — otherwise "no mutating webhook" is also true of an empty file and of a cluster where the operator was never deployed.
- The HTTP probes are gated on an `OGX answers at all` control, and the serves-Responses assertion contrasts against a probe of a path OGX certainly does not route.
- Both static guards were verified to fail, for the right reason, against a deliberately reintroduced mutating webhook and a deliberately added CRD default on `spec.praxisMode`.

## Proving `/v1/responses` is served

This can't be a 2xx assertion — a POST with no usable model is a legitimate error whatever the routing looks like. The operator disables an API by omitting it from the config's `apis:` list, so OGX registers no FastAPI route, and (`ogx-ai/ogx` `server.py` has no `StarletteHTTPException` handler) an unrouted path falls through to Starlette's default `404 {"detail": "Not Found"}`.

So the suite probes a path OGX certainly does not route, and requires `/v1/responses` not look like it. A `GET` corroborates through a different mechanism: a registered router makes the collection path known, so an unsupported method yields `405`, while an absent router yields the same plain `404`.

## Known gaps, stated rather than papered over

- The probes run **in-cluster**. Neither suite reaches OGX from genuinely outside; the external-path assertions are structural. A real external-reach test needs the downstream OCP/QE counterpart.
- kind runs kindnet, which creates NetworkPolicy objects and enforces nothing. Enforcement stays behind `OGX_E2E_TEST_NP_ENFORCEMENT` in `TestNetworkPolicySuite`.
- Route and HTTPRoute checks are vacuous where the CRDs are absent. The suite **logs this loudly** rather than reporting a silent pass.
- The opt-in `/v1/conversations` probe is skipped on ogx-ai/ogx#6558 — OGX force-adds `conversations` to the served set regardless of the `apis:` list. The operator's half (the generated config omits it) is still asserted; the skip is the regression test for the upstream fix.

## Also worth a decision

RHAISTRAT-2277's acceptance criterion — a single external `/v1/responses` route with port 8321 unreachable — describes the **opt-in** topology, not 3.6's default. `tests/e2e/praxis_optin_test.go` is what will prove it once it applies by default; as flagged on the ticket, the criterion itself needs revisiting or deferring past 3.6.

## Verification

`make lint`, `make manifests generate api-docs` (no diff), `hack/check_go_errors.py`, and `make test` all clean. The e2e suites need a cluster with the operator deployed.

### Rebase note

Rebasing onto `feature/praxis` (post-#373) broke one reference: `api/v1beta1/praxis_default_test.go`'s `TestIsPraxisModeEnabled_DefaultsToDisabled` called the old private `isPraxisModeEnabled()` helper, which #373 replaced with the public `Spec.IsPraxisModeEnabled()` method — and #373 also gave that method a different, deliberate default (writing `spec.praxisMode` at all, even empty, opts in with `enabled` defaulting to `true`, matching the CRD's structural-schema default). Updated the test to call the new method and corrected its two subtests that asserted the old (pre-#373) semantics; this now agrees with `TestCRDDoesNotDefaultPraxisMode` elsewhere in the same file, which already asserted the CRD-level default correctly. `go build`, `go vet`, and `make test` are green on the rebased branch.


eoinfennessy (approved):
I didn't look in great detail, but this generally lgtm, and seems to cover everything in the associated Jira.

VaishnaviHire (approved):

PR #378, opened 2026-09-21
fix: remove unrestricted SCC rule

VaishnaviHire:


VaishnaviHire:
@mergify rebase

rhdedgar (approved):
+1 to the changes if the e2e tests can pass.

mfleader (approved):
I tested this PR on OpenShift 4.21. The controller-manager pod runs under restricted-v2, and the operator still creates the nonroot-v2 RoleBinding.

PR #379, opened 2026-09-21
fix(config): emit persistence block for declarative pgvector provider (RHAIENG-6374)

mfleader:
[RHAIENG-6374](https://redhat.atlassian.net/browse/RHAIENG-6374)

## Description

`expandPgvectorProvider` omitted the `persistence` block from the generated pgvector provider config. The server dereferences `config.persistence` in `initialize`, so a declarative pgvector deployment crashed at startup with `AttributeError: 'NoneType' object has no attribute 'backend'`.

The fix emits `persistence` with `backend: kv_default` and a namespace from the new `PgvectorProvider.PersistenceNamespace()`:

- With an explicit `id`, the namespace is `vector_io::<id>`, so two named pgvector instances on one `OGXServer` do not share vector store records.
- With no `id`, the namespace is `vector_io::pgvector`, the value every shipped base config uses, so an install moving from a base config pgvector provider to the declarative path still reaches the records it already wrote.

One pair still collides: an entry with no `id` alongside an entry with `id: pgvector`. The webhook admits it because the provider ids differ. That is tracked as a follow-up.

## How has this been tested?

`make test`, `make test-e2e` and `make lint` pass on this branch, rebased onto `be51587`. `make test-e2e` ran on a kind cluster with cert-manager and the operator image built from this branch.

Unit tests in `pkg/config` cover the no-id and explicit-id namespaces, and a second test builds two pgvector providers in one CR and fails if they share a namespace. The configgen golden file covers the no-id case through generation. Reverting `pkg/config/provider.go` to `be51587` fails the unit tests with `expected persistence map, got <nil>`, and replacing the namespace with a constant fails the two-instance test.

On a kind cluster with PostgreSQL and the 3.6 EA2 server image:

| | operator at `be51587` | this branch |
|---|---|---|
| persistence in generated config | absent | present |
| server pod | CrashLoopBackOff, 5 restarts | Running, 0 restarts for 5 minutes |
| AttributeError in logs | 1 | 0 |

The server echoes the block back on `GET /v1/providers`, confirming it parsed and accepted what the operator emitted.

The test plan is in [opendatahub-test-plans pull request 20](https://github.com/opendatahub-io/opendatahub-test-plans/pull/20).


VaishnaviHire (approved):
Thanks for the fix!

